PR #3, opened 2019-08-28
Pytorch

yafshar:
@mjwen I have updated the c++ codes with the new STL-like container for multi-dimensional arrays. Underlying storage is a 1-dimensional array. I removed all the allocation and deallocation by hand to avoid unwanted exceptions and memory leak.

Also, I added a CDDL LICENSE file. 
I am not sure if this is the policy you want to follow. We can remove it.


mjwen:
@yafshar Thanks for the PR. It looks great, especially the STL-like container for multi-dimensional arrays. I will merge it. 

FYI, I tend to format the C++ source using "clang-format" (with the same style as KIM-API) and the python source using "black". This can be easily done for all sources via [utils/format_soruce.py](https://github.com/mjwen/kliff/blob/master/utils/format_sources.py). I will format it after merge.

PR #42, opened 2022-03-28
Fix to the parameter transformation functionality

yonatank93:
List of changes:
* Set `kliff.model.Models().model_params` as an attribute that store
the parameter values in the original parameterization.
* Move the transformation and inverse transform to
`kliff.model.OptimizingParameters()`, e.g., the transform function in `get_opt_params` and
the inverse transform in `update_opt_params` methods. This also requires the parameter
transfomation class to be passed in to `OptimizingParameters`.
* Add a line in the echo methods saying that the printed values are in
the original parameterization.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   lint:
--- a/kliff/models/kim.py
+++ b/kliff/models/kim.py
@@ -626,7 +626,7 @@ class KIMModel(Model):
                     )
 
         # When params_transform is set, a user can do whatever in it
-        # function, e.g. update params is not an optimizing parameter.
+        # function, e.g. update a parameter that is not an optimizing parameter.
         # In general, we do not know how parameters are modified in there,
         # and therefore, we need to update all params in model_params to kim
         # Note, `params_transform.inverse_transform()` is called in
--- a/kliff/models/model.py
+++ b/kliff/models/model.py
@@ -1,3 +1,4 @@
+import copy
 import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, TextIO, Tuple, Union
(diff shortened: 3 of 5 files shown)


mjwen:
@yonatank93 thank for the PR! Can you please enable `Allow edits from maintainers` for me to make some adjustment?

yonatank93:
@mjwen I enabled `Allow edits from maintainers`. Let me know if this is not working.

mjwen:
The functionality to deal with parameter transformation has all been moved to the `Model` class, leaving `OptimizingParams` free. Except the below behaviors: 

- `model.echo_opt_params()` will always show the transformed params 

- A new parameter `parameter_space` added `model.echo_model_params(parameter_space=...)`. A value of `original` will show params in the untransformed space (this is the default behavior), and a value of `transformed` will show it in the transformed space.   

@yonatank93 Can you test this out and let me know whether this solves your problem?

mjwen:
Agreed! It's useful to let the user know that it's in the transformed space. Added. 

Also, it would be helpful to add a tutorial example to demonstrate this. Can you convert your Jupyter notebook to an example as those listed [here](https://github.com/openkim/kliff/tree/master/examples)? Probably take a look at  [example_kim_SW_Si.py](https://github.com/openkim/kliff/blob/master/examples/example_kim_SW_Si.py). The docstrings in the examples follow the [sphinx-gallery](https://sphinx-gallery.github.io/stable/syntax.html) syntax.

yonatank93:
Sure, I will do it later today. In the example, we want to show:
* How to use, say, log transformation, which will also include some motivation why we want to transform the parameter space.
* Do we also want to show how to construct different parameter transformation class?
* Update the parameter values in the transformed space.
* Do we want to show the difference in the result when we use the original vs transformed space?

mjwen:
Thanks! 
- We may not want to show how to construct different parameter transformation in this tutorial. But can say that a user can do it easily by subclassing the `ParameterTransform` class. 

- It might be difficult/lengthy to compare the results. So, probably not do it at this stage. We may need longer docs/tutorials once all the UQ stuff is fleshed out. Then, we can do it in a more thorough way. 

I will merge and close this. Feel free to open a new PR for the tutorial.

PR #44, opened 2022-03-31
Add an example for parameter transformation

yonatank93:
The example was set up to run similar calculation as
``example_kim_SW_Si.py``, but done in log parameter space. This is
done by applying log parameter transformation to the model.
This example shows how to use parameter transformation. It also shows
that after setting the parameter transformation, further calculation
can be done in the same way as if we use the default, original space.


mjwen:
thanks, merged!

PR #47, opened 2022-04-11
Setting cutoff to a particular value

adityakavalur:
When setting cutoff. It should check for the second entry being fix, the first entry is float. Trying to lowercase a float object results in an error


mjwen (approved):

mjwen:
@adityakavalur Thanks for the PR! merged.

PR #48, opened 2022-04-20
WIP: Adding weight information to KLIFF

yonatank93:
Currently, I just added an option to pass in c1 and c2 via `weight_params` argument in Dataset and Configuration. The weights are calculated by $1/w = c1^2 + c2^2 * \Vert f \Vert^2$. The property `config.weight` is a dictionary containing the computed weights in this way. I modified the default residual function to use these weights.


kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+
+    def __init__(
+        self,
+        config_weight: float = 1.0,
+        energy_weight: float = 1.0,
+        forces_weight: float = 1.0,
+        stress_weight: float = 1.0,
+    ):
+        self._config_weight = config_weight  # weight for the entire configuration
```
mjwen, line 30: Can we remove the comment, since it's provided in the docs.

kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+default_weight_params = {
+    "energy_weight_params": [1.0, 0.0],
+    "forces_weight_params": [1.0, 0.0],
+    "stress_weight_params": [1.0, 0.0],
+}
+
+
+class NonuniformWeight(Weight):
+    r"""
+    Non-uniform weight that is computed from the data. The calculation follows Lenosky et
```
mjwen, line 65: Would be good to give a full reference at the end of the doc string.

kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def forces_weight(self):
+        return self._forces_weight
+
+    @property
+    def stress_weight(self):
+        return self._stress_weight
+
+
+default_weight_params = {
```
mjwen, line 56: Since this is only used in `NonuniformWeight`, I'd suggest we move the default_weight_params to be within the class. You can place it just above the `__init__` method, and get access to in anywhere in the class using self.default_weight_params

kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+        forces_weight: float = 1.0,
+        stress_weight: float = 1.0,
+    ):
+        self._config_weight = config_weight  # weight for the entire configuration
+        self._energy_weight = energy_weight
+        self._forces_weight = forces_weight
+        self._stress_weight = stress_weight
+
+    @staticmethod
+    def compute_weight(config):
```
mjwen, line 36: Would be better to add typing for it. `(config:Configuration)`. Same for the NonuniformWeight below.

kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+                    raise WeightError(
+                        f"Expect the keys of `weight_params` to be one or combinations "
+                        f"of {', '.join(default.keys())}; got {key}. "
+                    )
+                else:
+                    if (
+                        np.ndim(value) == 0
+                    ):  # If there is only a number given, use it to set c1
+                        default[key][0] = value
+                    else:  # To set c1 and c2, a list with 2 elements need to be passed in
```
mjwen, line 159: Maybe check shape = (2,) for `value` to make sure it's what we need?

kliff/dataset/weight.py
```diff
@@ -0,0 +1,167 @@
...
+
+        self._weight_params = self._check_weight_params(
+            weight_params, default_weight_params
+        )
+
+    def compute_weight(self, config):
+        """
+        Compute the weights of the energy, forces, and stress data.
+        """
+        energy = config._energy
```
mjwen, line 100: energy = config.energy should be good. Same for forces and stress.

kliff/loss.py
```diff
@@ -240,14 +191,11 @@ def __init__(
...
         default_residual_data = {
-            "energy_weight": 1.0,
-            "forces_weight": 1.0,
-            "stress_weight": 1.0,
             "normalize_by_natoms": True,
         }
 
         residual_data = _check_residual_data(residual_data, default_residual_data)
-        _check_compute_flag(calculator, residual_data)
+        # _check_compute_flag(calculator, residual_data)
```
mjwen, line 198: Can you update the `_check_comute_flag` function to something like  `_check_comute_flag(calculator, weight, residual_data)` and check weight there?

kliff/dataset/dataset.py
```diff
@@ -184,16 +195,45 @@ def stress(self) -> List[float]:
     @property
     def weight(self):
         """
-        Get the weight of the configuration if the loss function.
+        Get the weight class of the loss function.
         """
         return self._weight
 
+    @property
```
mjwen, line 202: We'd expect all kinds of weight be accessed via self.weight, instead of the config_weight, ... stress_weight properties. So, I guess the four properties can be deleted?

mjwen (commented):
Hi @yonatank93 nicely done, this is great! I just have a few minor comments you may want to consider. 

Also, can you add a test in the tests directory to simply test the NonuniformWeight class?

kliff/dataset/weight.py
yonatank93: I will do this. What do you think about the name of the class? I actually think that `NonuniformWeight` is not very descriptive. Do you have some suggestion for the name?

kliff/dataset/weight.py
yonatank93: I did this because if I call `config.stress` when the dataset doesn't contain stress, it raises an error. I can use try catch to overcome this.

kliff/loss.py
yonatank93: I commented out `_check_compute_flag` because I am not sure what to check in the residual data. Previously, it checked if the weights are big enough and put a message in the log file if the weights are small. So, when you suggest to update `_check_compute_weight`, is it to keep this functionality (to check if the weight are large enough and give feedback to user)?

yonatank93 (commented):
@mjwen There are several questions that I have about your comments.

kliff/dataset/weight.py
mjwen: Maybe `MagnitudeInverseWeight`, indicating that the weight is proportional to the inverse of the magnitude?

kliff/dataset/weight.py
mjwen: Ah, right. Then let's keep the underscore version for now. Will probably update the Configuration class later.

kliff/loss.py
mjwen: Yes, that's what I meant.

kliff/dataset/dataset.py
yonatank93: Does this mean if user wants to access the weight, he/she would do `config.weight.energy_weight`, etc.?

mjwen: Yes. I believe this is a bit cleaner, instead of attaching all the stuff to the Configuration class.

kliff/loss.py
yonatank93: So would something like the following inside `_check_compute_flag` do?
```
    ew = weight.energy_weight
    if calculator.use_energy and ew < 1e-12:
         logger.warning(msg.format("energy", ew))
```

mjwen:
Yes. Ecactly
________________________________

kliff/loss.py
yonatank93: @mjwen Is there a reason we want to keep the argument `residual_data` in `_check_compute_flag`? Because it seems like this function only checks the weights, and now `resudual_data` doesn't have information about the weights anymore.

kliff/dataset/weight.py
yonatank93: What I noticed is that if I do `config:Configuration`, where is an importing loop, since `Configuration` is in dataset.py and that script also imports `Weight`.

kliff/loss.py
mjwen: Good catch. You are right. I think we can remove `residual_data` from it now. 

Also, since we are using the new way to assign weight, the [doc](https://kliff.readthedocs.io/en/latest/modules/loss.html#doc-loss) is outdated as well. Can you update the doc as well? [Here](https://github.com/openkim/kliff/blob/master/docs/source/modules/loss.rst) is the source.

yonatank93: Sure, I can do that.

So, the `residual_data` pretty much will act like keyword arguments for the residual function then?

Since the primary purpose of `_check_compute_flag` is to check if the weights are large enough and give feedback to user about this, wouldn't it more appropriate to do this check when user creates the weight? That is, what if we move `_check_compute_flag` to weight class and call it inside `weight.compute_weight`?
I also think that this will be more future proof, for example if user wants to define some calculator that doesn't use energy, forces, or stress. In that way, they can define their own `_check_compute_flag` in their weight class.

mjwen: Yes, it is, residual_data is some additional args to residual_fn. Now that only normalize_by_num_atoms is left, I think we can redefine the signature the residual_fn as `residual_fn(prediction, reference, identifier,  weight, natoms, normalize_by_atoms)` to get rid of the use of `residual_data` in all places. Note, this may require the update of some tests and docs. 

> what if we move _check_compute_flag to weight class and call it inside weight.compute_weight?

This may not be easily doable, since it depends of the flags `compute_energy`, `compute_forces`, and `compute_stress`, which are defined in a `Calculator`. However, a `Configuration` is supposed to know nothing of the calculator.

yonatank93: > Yes, it is, residual_data is some additional args to residual_fn. Now that only normalize_by_num_atoms is left, I think we can redefine the signature the residual_fn as `residual_fn(prediction, reference, identifier, weight, natoms, normalize_by_atoms)` to get rid of the use of `residual_data` in all places. Note, this may require the update of some tests and docs.

I would still prefer to keep `residual_data`-like argument to be more flexible for the users, for example if users need to use some other information when they define their residual function. Maybe have the signature of the residual_fn as `residual_fn(prediction, reference, identifier, weight, natoms, **residual_data)`.

> This may not be easily doable, since it depends of the flags `compute_energy`, `compute_forces`, and `compute_stress`, which are defined in a `Calculator`. However, a `Configuration` is supposed to know nothing of the calculator.

For this, I was thinking something like the following as a check,
```
if config._forces is not None and np.all(fw < 1e-12):
    logger.warning(msg.format("forces", fw))
```
and suggest to set `use_energy=False` in `calc.create()`, even before users call `calc.create()`.
I have addressed most of your other comments and I can push my commit with an implementation on this weight check.

mjwen: > I would still prefer to keep residual_data-like argument to be more flexible for the users, for example if users need to > use some other information when they define their residual function. Maybe have the signature of the residual_fn as > residual_fn(prediction, reference, identifier, weight, natoms, **residual_data).

The default residual functions we provide can only deal with weights (now moved to the `weight` parameters) and whether to normalize by atoms. If a user wants to do more, a user defined residual_fn has to be provided and the user can do whatever needed. So, it seems not necessary to keep `residual_data`. But I agree that it seems better to keep `residual_data` to give a more controlled way of providing additional info. So, let's do what you've suggested. 

> ```
> if config._forces is not None and np.all(fw < 1e-12):
>      logger.warning(msg.format("forces", fw))
> ```
 
This may lead to some confusion, e.g. in the case where `calculator.use_forces=False`, then the value of `fw` does not matter at all, and this warning is not expected to be shown. But since our default fw is 1.0, hopefully a user playing with `calculator.use_forces` and `fw` will know what they are. I agree with that moving the check to `Configuration` is much cleaner. So, please go ahead and do it.    

> I have addressed most of your other comments and I can push my commit with an implementation on this weight check.

Great!

mjwen:
@yonatank93 let me know when it's ready for me to review

yonatank93:
@mjwen I have a question about typing in `Weight.compute_weight(config)`.
https://github.com/openkim/kliff/pull/48#discussion_r856893604

yonatank93:
@mjwen I am still working on updating the documentation for the loss function. I am also thinking of adding documentation about the weight in the [dataset](https://kliff.readthedocs.io/en/latest/modules/dataset.html) section.
I think you can review the other changes, although I still have the question about https://github.com/openkim/kliff/pull/48#discussion_r856893604.

kliff/dataset/weight.py
mjwen: hmmm, let's skip the typing for now.

yonatank93: Sounds good to me.

mjwen:
> @mjwen I am still working on updating the documentation for the loss function. 

No rushes at all! 

> I am also thinking of adding documentation about the weight in the [dataset](https://kliff.readthedocs.io/en/latest/modules/dataset.html) section.

Great! It would be good to refer the doc about weight added in dataset in the doc for loss function, since it is the place where the weight is actually used.

yonatank93:
@mjwen Please review the documentation I created/changed in KLIFF.

mjwen (approved):
@yonatank93 This is awesome! 

Thanks for reading through the docs and making modifications as necessary. I think we are good to merge. 

Are there more changes you want to make? If not, I'll merge it soon.

yonatank93:
@mjwen Thanks for catching some of the typos. Is an example for the weight necessary, on top of the documentation? If no, then I don't have anything to add/change.

yonatank93:
@mjwen I brought up this weight implementation in KLIFF diring KIM meeting today. There is a significant input, what if we retrieve the weight as a dictionary? That is, maybe we will have `weight.get_weight()` that returns
```
weight = {"energy_weight": [...], "forces_weight": [...], "stress_weight": [...]}
```
The thought is that it will be easier to extend the weight class if we have other material properties, aside from energy, forces, and stress.

mjwen:
> @mjwen Thanks for catching some of the typos. Is an example for the weight necessary, on top of the documentation? If no, then I don't have anything to add/change.

The current examples already used the `Weight` class, although not explained in great detail. I believe this is good enough given that we have nice documentation for `Weight` class.

mjwen:
> @mjwen I brought up this weight implementation in KLIFF diring KIM meeting today. There is a significant input, what if we retrieve the weight as a dictionary? That is, maybe we will have `weight.get_weight()` that returns
> 
> ```
> weight = {"energy_weight": [...], "forces_weight": [...], "stress_weight": [...]}
> ```
> 
> The thought is that it will be easier to extend the weight class if we have other material properties, aside from energy, forces, and stress.

I think our current approach should work better, because 

- the need to fit to other properties is really small; not so many people these days are fitting to other properties
- all the KLIFF framework is coded up using energy/forces/stress as reference data. If a user wants the fit to new properties, all other parts needs to be modified. Then, it's fine to modify Weight as well. 
- our `weight.energy_weight` class properties approach is cleaner than returning a gigantic dict. I personally prefer simplicity and better easier documentation than extensibility. If a user really wants to fit to other properties, the `Weight` class can be subclassed and new @property can be defined.

yonatank93:
Ok. I will keep this in mind as an alternative way to if we need update the weight in the future.

There is another issue I want to bring up.
For this update, I deleted the built-in residual functions, and only keep `energy_forces_residual`. My thought was that the `forces_residual` and `energy_residual` only set the respective weight to zero and use `energy_forces_residual`. So, the same action could also be done by setting the weight in `Weight` directly.
However, suppose that when creating calculator, I specify `use_forces=False`, hoping to speed up the calculation. But it seems that the built-in residual function doesn't know if we want to use certain property or not, because it always assumes that the first element of the prediction array is always energy and the rest to be forces.
So, should we either:
* Update the built-in residual function? Maybe append the key `use_forces`, `use_energy`, etc to `residual_data`? or
* Create other built-in residual functions?

mjwen:
>  However, suppose that when creating calculator, I specify `use_forces=False`, hoping to speed up the calculation. But it seems that the built-in residual function doesn't know if we want to use certain property or not, because it always assumes that the first element of the prediction array is always energy and the rest to be forces. So, should we either:
> 
> * Update the built-in residual function? Maybe append the `use_forces`, `use_energy`, etc from the calculator to `residual_data`? or
> * Create other built-in residual functions?

Good catch! The previous `energy_residual` and `force_residual` was not correct. Depending on the `use_energy` and `use_forces` flag, the prediction may be different. In retrospect, [this design](https://github.com/openkim/kliff/blob/d1209feac98387e44145644272fef85a788b1683/kliff/models/model.py#L113) is bad. We will need to modify this in the next version. 

But for now, I think let's create separate residual functions for different cases (your second option), just to ensure the the interface of a residual function is the same. Then [this part](https://github.com/openkim/kliff/blob/d1209feac98387e44145644272fef85a788b1683/kliff/loss.py#L255) needs to be be updated well: depending on the `use_energy`... flags, we need to choose a different residual function.

yonatank93:
I think in KLIFF, we can specify `use_energy`... for each configuration separately. If we update [this part](https://github.com/openkim/kliff/blob/d1209feac98387e44145644272fef85a788b1683/kliff/loss.py#L255) with if else statement, it seems that we assume that `use_energy`... is the same for all configurations. So, are we going to ignore this for now?

mjwen:
Technically, `use_energy` can be specified for each config separately. But we do not expose any user interface for doing that. More specifically, a single `use_energy` is provided by a user via a calculator, and this single value will be applied to all the configurations. So, it's safe to do what's mentioned above. 

If a user does not want to use energy for a configuration, this has to be achieved via the `energy_weight` through the new `Weight` class.

mjwen:
All looks good; merged. Thanks!

PR #55, opened 2022-05-26
WIP: Implementing MCMC sampling to KLIFF

yonatank93:
We want to integrate UQ analysis into KLIFF. In this implementation, we will implement MCMC via `ptemcee` and `emcee` Python packages. We also want to add some functions for the analysis of the samples. What's included:
* `uq.MCMC` class as a universal class to construct MCMC sampler and run the MCMC simulation.
* `uq.mser` function that estimates the burn-in time of the MCMC simulation.
* `uq.autocorr` function, which is a wrapper to `emcee.autocorr.integrated_time()` to estimate the autocorrelation length.
* `uq.rhat` function that computes the potential scale reduction factor (PSRF) as a convergence assessment tool.


mjwen (commented):
Thanks @yonatank93 for the PR! This looks great and I only have very minor comments. It would be great to: 
- add tests 
- add a tutorial example, probably pulling some stuff from the paper you are preparing?

yonatank93:
@mjwen I added tests and an example for the UQ implementation. Can you take a look at it?
I am also thinking of adding documentation under the "Frequently Used Modules" section, but haven't done it yet.

docs/source/auto_examples/example_uq_mcmc.py
```diff
@@ -0,0 +1,229 @@
...
+#    It has been shown that including temperatures higher than :math:`T_0` helps the
+#    convergence of walkers sampled at :math:`T_0`
+
+
+# Create sampler
+sampler = MCMC(loss, ntemps=4, threads=4, random=np.random.RandomState(seed))
+
+# Initial starting point. This should be provided by the user.
+bounds = calc.get_opt_params_bounds()
+p0 = np.empty((4, 4, 2))
```
mjwen, line 130: what are 4, 4, 2? I guess one is `ntemps`. Better use the variable names like `ntemps`

docs/source/auto_examples/example_uq_mcmc.py
```diff
@@ -0,0 +1,229 @@
...
+#
+# .. note::
+#    It has been shown that including temperatures higher than :math:`T_0` helps the
+#    convergence of walkers sampled at :math:`T_0`
+
+
+# Create sampler
+sampler = MCMC(loss, ntemps=4, threads=4, random=np.random.RandomState(seed))
+
+# Initial starting point. This should be provided by the user.
```
mjwen, line 128: How about setting the parameter bounds when defining them? Then a user can know how to do it.

docs/source/auto_examples/example_uq_mcmc.py
mjwen: Oh, I see this in the `note` below.

docs/source/auto_examples/example_uq_mcmc.py
```diff
@@ -0,0 +1,229 @@
...
+# Run MCMC
+sampler.run_mcmc(p0, 10000)
+
+# Retrieve the chain
+chain = sampler.chain
+
+
+#########################################################################################
+# .. note::
+#    The initial states need to be an array with shape ``(K, L, N,)``, where ``K`` is the
```
mjwen, line 143: "The initial states $p_0$" to be more specific.

docs/source/auto_examples/example_uq_mcmc.py
```diff
@@ -0,0 +1,229 @@
...
+#    number of temperatures, ``L`` is the number of walkers, and ``N`` is the number of
+#    parameters. As a default, ``M`` is set to be twice of ``N``, but we can also specify
```
mjwen, line 145: What is ``M``?  From below I see it is the steps. Let's define it here.

examples/example_uq_mcmc.py
```diff
@@ -0,0 +1,229 @@
+"""
```
mjwen, line 1: I put my comments for this file at auto_examples.

mjwen (commented):
My first impression of the tutorial example is that it is too complicated and it is a bit difficult for me to understand what's going on. This is probably because I am not extremely familiar with the UQ steps. If you can add more explanation of the `uq` package in the frequently used module as you planned, it could be helpful to get a better understanding.

docs/source/auto_examples/example_uq_mcmc.py
yonatank93: I actually agree, this is not clear. I have fixed the notation.

docs/source/auto_examples/example_uq_mcmc.py
yonatank93: I am not really sure how to make this a clearer process. How about we specify the prior bounds with `logprior_args` when instantiating `uq.MCMC` and define the variable `bounds` there? This is a redundant process, since I set it up so that when user doesn't give the bound of the uniform prior, it will extract it from the calculator.

docs/source/auto_examples/example_uq_mcmc.py
yonatank93: I like this suggestion.

docs/source/auto_examples/example_uq_mcmc.py
yonatank93: It should be ``L``. I thought to reserve ``M`` for the number of MCMC steps.

examples/example_uq_mcmc.py
yonatank93: I addressed the comments by updating examples/example_uq_mcmc.py in the next commit.

yonatank93:
I have reworked the examples. The next item on my list is to write the frequently used module section.

mjwen:
@yonatank93 The UQ module explanation looks great, thanks! I just have a comment regarding the bounds. Other than that, the PR is ready for merge.

yonatank93:
@mjwen What question do you have about the bounds?

mjwen:
> @mjwen What question do you have about the bounds?

Are the bounds extremely important parameters for the MCMC algorithm to work? In other words, you set them to (-8, 8) here; what if we do not set it and use the default value of logprior_args? If this does not hugely affect the result, I think we can removing providing initial bounds here and use the default values in logprior_args. If not, the current implementation looks fine to me.

for [docs/source/auto_examples/example_uq_mcmc.py](https://github.com/openkim/kliff/pull/55/files/c827ec536522172c27a81b3d9f6634d03d060d76#diff-d87ea5a2a9fccb44a2ec969f705fccdb7a8704f1a1d72fc655696e6b76722a66)

yonatank93:
> Are the bounds extremely important parameters for the MCMC algorithm to work? In other words, you set them to (-8, 8) here; what if we do not set it and use the default value of logprior_args? If this does not hugely affect the result, I think we can removing providing initial bounds here and use the default values in logprior_args. If not, the current implementation looks fine to me.

The variable `bounds` are used twice: (1) in setting the random initial positions of the walkers and (2) in setting the boundaries of the uniform prior. For the first case, the user can actually just set his/her own random initial positions. For the second use case, specifying the bounds (via `logprior_args`) is actually optional. When the user doesn't specify the bounds, then it will just retrieve it from the parameters.

Suppose if we didn't define the bounds and set the initial positions to be `np.random.uniform(low=-8, high=-8, ...)` and use the default values of `logprior_args` (which is actually also [-8, 8] in this example). I feel like the choice of the lower and upper bounds in `np.random.uniform` seem to be arbitrary.

There is an idea that I didn't explicitly mention in the example. It is suggested that when we run MCMC, we should set the initial positions of the walkers to be spread out in parameter space. In this example, the initial positions are pretty much points sampled from the prior. And I made this choice because the region in parameter space that has non-zero posterior probability is given by the region "inside" the uniform prior. So, I kind of want to communicate an idea on how to pick the initial positions.

mjwen:
Got it, thanks for the clarification! This is fine then.

Issue #62, opened 2022-07-12
I got error when passing in multiprocessing pool to `kliff.uq.MCMC`


yonatank93:
When we instantiate `kliff.uq.MCMC`, it instantiates sampler from `ptemcee` (`ptemcee.Sampler`) or `emcee` (`emcee.EnsembleSampler`). In both emcee and ptemcee, we can pass it additional keyword arguments, such as the random state and multiprocessing pool, and I set it so that `kliff.uq.MCMC` can do the same. (Pretty much, I passed in the entire kwargs to either `ptemcee.Sampler` or `emcee.EnsembleSampler`).

However, we have a problem when we declare the multiprocessing pool before instantiating `kliff.uq.MCMC`. The error message I got was:
```AttributeError: Can't get attribute 'loglikelihood_fn' on <module 'kliff.uq.mcmc' from 'kliff/kliff/uq/mcmc.py'>```
(See this [example](https://gitlab.com/yonatank93/kliff_uq/-/blob/master/examples/example_ptemcee.py) to reproduce this issue)

What I understood from this [discussion](https://stackoverflow.com/questions/36533134/cant-get-attribute-abc-on-module-main-from-abc-h-py), this error is because the `loglikelihood_fn` in `kliff.uq.MCMC`, which is used in the parallel run,  is defined after I declare the pool (see these [lines](https://github.com/openkim/kliff/blob/master/kliff/uq/mcmc.py#L200-L224)).

As a work around, I was thinking about adding a property `pool` with its setter to set the multiprocessing pool for the sampler, and force users to set the pool after instantiating `kliff.uq.MCMC`. For example,
```python
sampler = kliff.uq.MCMC(...)
sampler.pool = multiprocessing.Pool(nprocesses)
```

Do you have any suggestion on how to deal with this issue?

mjwen:
I ran the example you've linked, but got different types of error: 
```
  raise RuntimeError('''
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

I am running on Mac, are you running on Linux?

yonatank93:
[...]

PR #66, opened 2022-08-02
Fixing uq pool

yonatank93:
## Summary

Update the `uq` module to behave better when we run the sampling in parallel, especially when we use hybrid parallelization using multithreading in loss evaluation and MPI in sampling.

This is related to Issue #62 .

## Known Issue

* Typically, I would declare multiprocessing pool before instantiating the sampler. However, in the previous version, if we do this then we will get an error. What I understood was that it is because we need to declare the pool after we define all the functions to use in parallel. The problem in our UQ implementation is that the likelihood and/or posterior are defined inside `kliff.uq.MCMC().__init__()`, which I think is the reason why we would get an error if we declare the pool before instantiating `kliff.uq.MCMC`.
* In the previous version, when we use multithreading for the loss evaluation and MPI for the sampling (e.g., via `schwimmbad.MPIPool`, the performance was similar to if we use serial calculation in the sampling process (at least this was my experience).

## Fix

* I made the module to force user to declare the multiprocessing pool after instantiating `kliff.uq.MCMC`.
* I found that if `kliff.uq.PtemceeSampler` inherits from `ptemcee.Sampler` (also applies with the emcee counterpart), then this parallel computing performance issue seemed to disappear. I am not sure why this is the case, but at least it kind of fix the issue.
* I also updated the tests and documentation.


yonatank93:
@mjwen Can you take a look at this? And do we want to add a documentation in "How To: Run in parallel mode" about possible parallelization when running MCMC sampling? For example, by showing what user can do to use multithreading in loss evaluation and MPI in sampling, and point out that currently we don't support MPI in loss evaluation when running MCMC sampling.

mjwen:
Thank you @yonatank93 for the PR! I will take a look at this soon. In the meantime, can you fix to let the linting and test pass? For the test, it seems `ptemcee` is needed, but not installed in the GH actions. You may want to install it, like, below [this line](https://github.com/openkim/kliff/blob/9f338562429885758c701a7aa8e4989511b967c0/.github/workflows/testing.yml#L64).

kliff/uq/mcmc.py
```diff
@@ -106,6 +106,11 @@ def __new__(
         **kwargs,
     ):
 
+        if "pool" in kwargs:
+            raise TypeError(
```
mjwen, line 110: Change `TypeError` to `ValueError`?

kliff/uq/mcmc.py
```diff
@@ -135,237 +140,218 @@ def __new__(
...
+                self.T = get_T0(self.loss)
+            else:
+                self.T = T
+            logl_fn = self._loglikelihood_wrapper
+            logp_fn = self._logprior_wrapper(logprior_fn, *logprior_args)
 
-        self.sampler = emcee.EnsembleSampler(
-            nwalkers, ndim, logprobability_fn, **kwargs
-        )
+            global logprobability_fn
```
mjwen, line 329: Do we have to use global function? If possible, can this function be defined outside the class and used here? 

Also, it worries me that we have two global functions with the same name `logprobability_fn` here and above. Will they clash with each other?

kliff/uq/mcmc.py
```diff
@@ -106,6 +106,11 @@ def __new__(
         **kwargs,
```
mjwen, line 106: How about change `use_ptsampler` to `sampler`, which accepts a string as argument, that default to `pt_sampler`? Then if a user provides `ensemble_sampler`, it selects the other one. 

By doing this, we can easily extend the code for other samplers maybe in the future. We can even accepting an instance of a sampler class instead of a string.

mjwen:
> I have updated so that the linting and test pass. And coming back to my question, do we want to provide an example/discussion about how to run multiprocessing for UQ?

Yes, I think this would be great! This would be as similar as the existing UQ example with changes only reflecting the part.

kliff/uq/mcmc.py
yonatank93: I like this suggestion. Please see my updates in bbc37d1. I also updated the example and documentations, but I haven't included the discussion about parallelization.

kliff/uq/mcmc.py
yonatank93: I use global function because otherwise I always got the following error:
```
TypeError: cannot pickle 'kimpy.model.PyModel' object
```
Do you have any suggestion to fix this?

yonatank93: @mjwen The `PtemceeSampler` class uses `loglikelihood_fn` and the `EmceeSampler` class uses `logprobability_fn`. So they won't clash with each other.

And from what I tried, we can avoid declaring a global variable if we define `logprobability_fn` before instantiating `kliff.uq.MCMC`, and pass in that function as an argument of this class. However, it seems to defeat the purpose of this package if we require user to define this function manually. Instead, what if we add an optional argument to `kliff.uq.MCMC` to allow user to pass in their own `logprobability_fn` and have a default value with an if-else statement like the following:
```
if logprobability_fn is None:
    global logprob_fn
    def logprob_fn:
        ...
else:
    logprob_fn = logprobability_fn
```
Unfortunately, I still don't know how we can get rid of the global variable completely.

mjwen:
@yonatank93 You probably did not see my previous reply. The PR is ready to be merged, any last changes you want to make?

yonatank93:
@mjwen Actually, I have not worked on the documentation about parallelization for the MCMC sampling yet. So, do we want to add this documentation before merging this PR?

I also haven't figured out how to avoid using global variable. I want to work on this, but it can be for another PR.

yonatank93:
@mjwen I just added a more thorough documentation about parallelization in MCMC. Do you have any comments on that? Otherwise, I don't have any other changes for this PR.

mjwen (approved):
Looks all good to me! Merged.

PR #107, opened 2023-03-14
WIP: Implement bootstrap

yonatank93:
We want to implement a bootstrap method of UQ. In general, we do this by taking the compute arguments and samples from the list of compute arguments with replacement. Then, we train the model using this sample of compute arguments. The optimal parameters give a point in the ensemble.

TODO:
- [x] Documentation
- [x] Example(s)
- [x] Tests
- [x] Discuss changes done in some internal functions


yonatank93:
Note that this is still a very rough draft, though functional.
The intended use of the class, at least for empirical models, is illustrated in [examples/examples_bootstrap_SW_Si.py](https://github.com/yonatank93/kliff/blob/implement_bootstrap/examples/examples_bootstrap_SW_Si.py).
I want to have a similar workflow/syntax for NN models.

mjwen:
thanks @yonatank93! Let me know when ready and I'll take a look!

yonatank93:
I will @mjwen. Sorry that it is still very premature. By the way, the [example](https://github.com/yonatank93/kliff/blob/implement_bootstrap/examples/examples_bootstrap_SW_Si.py) for running bootstrap that I currently have reflects the workflow I intend users to use. That is, I want users to get an ensemble of parameters first before propagating the error to some other predictions, or in other words, the set of parameters is fixed. This is to answer your previous email.

mjwen:
@yonatank93 totally ok! I just wanted to make sure I get notified when it is ready.

yonatank93:
@mjwen The bootstrap implementation is ready for you to look at. I haven't added anything in `kliff/doc`, I will work on this once we are ok with this implementation.

kliff/calculators/calculator.py
```diff
@@ -364,6 +372,17 @@ def get_calculator_list(self):
             calc_list.extend([calc] * N)
         return calc_list
 
+    def has_opt_params_bounds(self):
+        """
+        Return a boolean to indicate whether there are parameters whose bounds are
+        provided.
+        """
+        calc_list = self.get_calculator_list()
```
mjwen, line 380: Can directly use3 `self._calculators`. This can make it a bit faster.

`self.get_calculator_list()` is nothing but repeat `self._calculators` multiple times.

kliff/calculators/calculator_torch.py
```diff
@@ -253,6 +276,89 @@ def _compute_stress(denergy_dzeta, dzetadr, volume):
...
+        Convert the parameters from a 1D array format to nested lists format.
+        """
+        sizes, nelems, _ = self.get_size_opt_params()
+        # Indices to index the flat array to get the appropriate portion of each parameter
+        # tensor
+        idx = np.append(0.0, np.cumsum(nelems)).astype(int)
+        parameters = []
+        for ii, size in enumerate(sizes):
+            params = flat_params[idx[ii] : idx[ii + 1]]
+            parameters.append(torch.Tensor(params.reshape(size)))
```
mjwen, line 359: The parameters are going to be used only at reference time, i.e. for predictions? right? Then this is probably OK. 

But I still recommend replacing `torch.Tensor` by `torch.Parameter` https://pytorch.org/docs/stable/generated/torch.nn.parameter.Parameter.html. The latter is a special tensor that requires grad, which is necessary of we want to take automatic differentiation w.r.t. the parameters.

kliff/calculators/calculator_torch.py
```diff
@@ -121,18 +122,40 @@ def create(
...
             )
 
+        # Finally, assign fingerprints dataset property as a FingerprintsDataset instance
+        self.fingerprints_dataset = FingerprintsDataset(self.fingerprints_path)
+
+    def get_fingerprints_dataset(self):
+        """
+        Return a list of fingerprints of the configurations.
+        """
+        return self.fingerprints_dataset.fp
```
mjwen, line 132: The function name and the return are inconsistent. Can we rename the function to `get_fingerprints`?

kliff/calculators/calculator_torch.py
```diff
@@ -121,18 +122,40 @@ def create(
...
         loader = DataLoader(
-            dataset=fp, batch_size=batch_size, collate_fn=fingerprints_collate_fn
+            dataset=self.fingerprints_dataset,
+            batch_size=batch_size,
+            collate_fn=fingerprints_collate_fn,
         )
 
         return loader
 
+    def set_compute_arguments(self, fingerprints):
```
mjwen, line 146: Can we rename this to `set_fingerprints`, as the inverse of the above function?

kliff/uq/bootstrap.py
```diff
@@ -0,0 +1,719 @@
+"""I am still working on this script to do bootstrap UQ."""
```
mjwen, line 1: Needs to be updated.

kliff/uq/bootstrap.py
```diff
@@ -0,0 +1,719 @@
...
+        Parameters
+        ----------
+        filename: str or Path
+            Where to export the bootstrap compute arguments samples
+        """
+        self._save_bootstrap_compute_arguments(
+            filename, get_identifiers_from_compute_arguments
+        )
+
+    def load_bootstrap_compute_arguments(self, filename: str):
```
mjwen, line 320: Is it important to know what configs are in each bootstrapped dataset? If not, in `save_bootstrap_compute_arguments`, the cas can be dumped as a pickle, instead of json? 

If it is important, we can probably write two files, one is what you are doing now, the other is the pickle. Then when loading them back, we can simply check their identifiers agree with each other. This avoids creating the compute arguments when loading them back. 

A third option, the identifiers and real compute arguments can be easily combined in a data structure, say a dictionary, and save to pickle. But I prefer the second approach above, in case our code base changes, we can still read from a simple json file what are the bootstrapped datasets.

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,133 @@
+"""
+.. _tut_mcmc:
```
mjwen, line 2: Name to be updated?

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,133 @@
...
+from kliff.utils import download_dataset
+
+##########################################################################################
+# Before running bootstrap, we need to define a loss function and train the model. More
+# detail information about this step can be found in :ref:`tut_kim_sw` and
+# :ref:`tut_params_transform`.
+
+# Instantiate a transformation class to do the log parameter transform
+param_names = ["A", "lambda"]
+params_transform = LogParameterTransform(param_names)
```
mjwen, line 48: Let's remove the `params_transform`. This adds additional barriers, and it is not the point of this tutorial.

mjwen (changes requested):

kliff/calculators/calculator_torch.py
yonatank93: Currently, I only need this to retrieve the parameters to export the samples, and when we want to use the samples to update the model parameters and use the model for other calculations. But, I can see that your suggestion is more future-proof.

yonatank93:
Hi @mjwen, I have applied the minor feedback that you have. There are several things that I want to discuss with you.
* Currently I have `bootstrap_compute_arguments` for both `BootstrapEmpiricalModel` and `BootstrapNeuralNetworkModel`. Especially for neural network model, we don't use compute arguments, but we use the fingerprints to generate the bootstrap dataset. Do you have any preference if we should keep it as is now or if we should call it `bootstrap_compute_arguments` for empirical model and `bootstrap_fingerprints` for neural network model? Another option is to call both as `bootstrap_dataset`. What's your thought?
* In your [comment](https://github.com/openkim/kliff/pull/107#discussion_r1143602553), I am not getting your point about creating compute arguments. I think I decided to export the bootstrap compute arguments as a json file containing the identifiers so that I can just export as little information needed as possible. I assume when we instantiate `Loss`, which is a required argument for `Bootstrap`, we generate the compute arguments. In the `load_bootstrap_compute_arguments`, my intention is to use this dictionary of identifiers and the compute arguments in `Loss` to generate bootstrap compute arguments by the means of querying. Please let me know your thought on this.

mjwen:
> Hi @mjwen, I have applied the minor feedback that you have. There are several things that I want to discuss with you.
> 
> * Currently I have `bootstrap_compute_arguments` for both `BootstrapEmpiricalModel` and `BootstrapNeuralNetworkModel`. Especially for neural network model, we don't use compute arguments, but we use the fingerprints to generate the bootstrap dataset. Do you have any preference if we should keep it as is now or if we should call it `bootstrap_compute_arguments` for empirical model and `bootstrap_fingerprints` for neural network model? Another option is to call both as `bootstrap_dataset`. What's your thought?

Let's keep what we have now? Even for the original neural networks, we do not have compute_arguments (in the sense of kim models), but fingerprints. We specifically name the fingerprints as compute arguments to keep a uniform interface for the optimizer. In retrospect, it might not be a good idea, but let's keep it consistent for now. 

> * In your [comment](https://github.com/openkim/kliff/pull/107#discussion_r1143602553), I am not getting your point about creating compute arguments. I think I decided to export the bootstrap compute arguments as a json file containing the identifiers so that I can just export as little information needed as possible. I assume when we instantiate `Loss`, which is a required argument for `Bootstrap`, we generate the compute arguments. In the `load_bootstrap_compute_arguments`, my intention is to use this dictionary of identifiers and the compute arguments in `Loss` to generate bootstrap compute arguments by the means of querying. Please let me know your thought on this.

You are right. I misunderstood. I was thinking you read in the bootstrapped identified and recreate the compute arguments, which is not ideal. But in fact, you are only using it to index the compute arguments. This is all good. 

A minor stuff: for docs, we've transferred from the numpy style to the google style (see [here](https://github.com/openkim/kliff/blob/68db8956f2b38bc0846ff37cf4b72832278066f5/kliff/models/parameter.py#L16) for an example), so can you please update yours? It has some minor benefits: because we are using type hints now, we do not need to write the type in the doc string, and we have used a package to parse it from the type hints and add the type to the generated html docs.

yonatank93:
@mjwen I have updated the documentation to use google style. Hopefully, I didn't miss any. Another thing that I am planning to work on is writing a documentation page about bootstrapping, similar to what we have for MCMC.

yonatank93:
Hi @mjwen. I just added the documentation page for bootstrapping.

kliff/calculators/calculator_torch.py
yonatank93: I changed `torch. Tensor` to `torch. Parameter`, but I got some unwanted behavior in `update_model_parameters`. When I changed to use `torch. Parameter`, I changed the line `param.data = parameters[ii]` to `param = parameters[ii]`, where `parameters[ii]` now is `torch. Parameter`. However, from my testing, this doesn't change the parameters. I think the way to update the parameters should be by using `param.data = <torch. Tensor>`. Because of this, I will revert back the change. I will also add a test for this functionality.

kliff/calculators/calculator_torch.py
```diff
@@ -253,6 +276,89 @@ def _compute_stress(denergy_dzeta, dzetadr, volume):
...
+        Parameters
+        ----------
+        parameters: np.ndarray
+            New parameter values to set. It needs to be a 1D array.
+        """
+        # Convert to the right format
+        parameters = self._convert_from_flat_parameters(parameters)
+        # Update the weights and biases
+        for ii, param in enumerate(self.model.parameters()):
+            param.data = parameters[ii]
```
mjwen, line 334: This is alright! I missed this previously.

kliff/calculators/calculator_torch.py
mjwen: My bad. What you are doing is 100% correct. I missed that you are assigning this back to `tensor.data` in my new comment above. Sorry about that!

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from kliff.calculators import Calculator
+from kliff.dataset import Dataset
+from kliff.loss import Loss
+from kliff.models import KIMModel
+from kliff.models.parameter_transform import LogParameterTransform
```
mjwen, line 35: `LogParameterTransform` can be removed?

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,131 @@
...
+from kliff.uq.bootstrap import BootstrapEmpiricalModel
+from kliff.utils import download_dataset
+
+##########################################################################################
+# Before running bootstrap, we need to define a loss function and train the model. More
+# detail information about this step can be found in :ref:`tut_kim_sw` and
+# :ref:`tut_params_transform`.
+
+# Instantiate a transformation class to do the log parameter transform
+param_names = ["A", "lambda"]
```
mjwen, line 45: Remove the above two lines, add add `param_names = list(opt_params.keys())` below when it is needed?

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,131 @@
...
+##########################################################################################
+# To perform UQ by bootstrapping, the general workflow starts by instantiating
+# :class:`~kliff.uq.bootstrap.BootstrapEmpiricalModel`, or
+# :class:`~kliff.uq.bootstrap.BootstrapNeuralNetworkModel` if using a neural network
+# potential.
+
+
+# It is a good practice to specify the random seed to use in the calculation to generate
+# a reproducible simulation.
+np.random.seed(1717)
```
mjwen, line 85: I don't love this way to specific seed. Can we implement it as `BootstrapEmpiricalModel(..., seed=<default>)` and provide a default?

examples/example_uq_bootstrap.py
```diff
@@ -0,0 +1,131 @@
...
+# Only use the forces data
+ca = calc.create(configs, use_energy=False, use_forces=True)
+
+# Instantiate the loss function
+residual_data = {"normalize_by_natoms": False}
+loss = Loss(calc, residual_data=residual_data)
+
+# Train the model
+min_kwargs = dict(method="lm")  # Optimizer setting
+loss.minimize(**min_kwargs)
```
mjwen, line 73: We don't necessarily need to train against the entire dataset before bootstrapping. It is ok to do this, but typical bootstrapping does not do this, right? 

This can actually lead to data leakage because the model is first trained on the entire dataset.

kliff/uq/mcmc_utils.py
```diff
@@ -68,18 +61,13 @@ def mser(
...
-        (nsteps, nwalkers, ndim). Note that the burn-in time needs to be discarded prior
-        to this calculation
-    args, kwargs
-        Additional positional and keyword arguments of ``emcee.autocorr.integrated_time``.
-
-    Returns
-    -------
-    float or array:
+    Args:
+        chain: Chains from the MCMC simulation. The shape of the chains needs to be
```
mjwen, line 65: For `np.ndarray` type, it still is good to add what they mean. Like it is here, `(nwalkers, nsteps, ndim,)` is very informative. The automated tool we use can only tell it is an np.ndarray. 

So, can you please add these back -- here and for other np.ndarray cases?

docs/source/modules/uq.rst
```diff
@@ -11,12 +11,28 @@ potential, to predict material properties that are not used in the
 training process. Thus, UQ process is especially important to assess the reliability of
 these out-of-sample predictions.
 
-In UQ process, we first quantify the uncertainty of the model parameters. Having found
-the parametric uncertainty, then we can propagate the uncertainty of the parameters and
-get the uncertainty of the material properties of interest, e.g., by evaluating the
-ensemble that is obtained from sampling the distribution of the parameters. As the first
-uncertainty propagation is more involved, KLIFF implements tools to quantify the
-uncertainty of the parameters.
+.. figure:: ../img/uq_cartoon.png
```
mjwen, line 14: Love the figure.

docs/source/modules/uq.rst
```diff
@@ -272,3 +286,124 @@ satisfied.
...
    See the tutorial for running MCMC in :ref:`tut_mcmc`.
+
+
+
+Bootstrap
+=========
+
+In general, the training dataset contains some random noise. When the data collection
+process is repeated, we will not get the exactly same values, but instead we will get
+(slightly) different values, where the diviation comes from the random noise. If we
```
mjwen, line 297: `diviation` ->`deviation`

docs/source/modules/uq.rst
```diff
@@ -272,3 +286,124 @@ satisfied.
...
+train the model to fit different realizations of the training dataset, we will get a
+distribution of the parameters. The uncertainty of the parameters from this distribution
+gives how the error in the training data is propagated to the uncertainty of the
+parameters. However, often times we don't have the luxury to repeat the data collection.
```
mjwen, line 301: `often times` -> `oftentimes`

docs/source/modules/uq.rst
```diff
@@ -272,3 +286,124 @@ satisfied.
...
+A suggestion in this case is to generate artificial datasets and train the model to fit
+these artificial datasets.
+
+.. figure:: ../img/bootstrap.png
+   :alt: Illustration on how bootstrapping works.
+
+Bootstrapping is a way to generate the artificial datasets. We assume that the original
+dataset contains :math:`N` *independent and identically distributed (iid)* data points.
+An artificial, bootstrap dataset is generated by sample :math:`N` points from the original
```
mjwen, line 310: Can we change `N` here to `M`? It does not need to be the same size as the original dataset.

mjwen:
@yonatank93 thanks! We are almost there. Can you please address the comments?

examples/example_uq_bootstrap.py
yonatank93: Is the concern here that the class can affect the numpy-wide generator? In that case, we can create a local random state and provide and make this an optional argument that user can specify. Any thoughts?

yonatank93:
@mjwen Thanks for the feedback. I have addressed most of them. There are a couple that I want to discuss with you:
* https://github.com/openkim/kliff/pull/107#discussion_r1150005677 
* https://github.com/openkim/kliff/pull/107#discussion_r1150008024

examples/example_uq_bootstrap.py
mjwen: I was thinking a bit simpler than that -- just moving `np.random.seed(<seed>)` into the `BootstripEmpiricalModel` class. 

It is more cosmetic than functional.

examples/example_uq_bootstrap.py
mjwen: Let's first make sure we are on the same footing. Below is a general way to do bootstrapping:
1. sample training set with replacement 
2. train the model with the sampled data 
3. and redo 1 and 2 multiple times 
4. evaluate on a holdout test set (mean and stdev mostly)

You see here we do not have the training on entire training set step. It does not have anything to do with running time, but how we need to do it. My argument was that we don't need to do the training on the entire training set step. It might be ok to do it, but it is then not the typical bootstrapping approach. 

Also, I don't think, in principle, the optimizations in different runs need to use the same optimizing hyerparams, although it is typically the case.

mjwen:
@yonatank93 sorry for the latest response. I just added comments to your two questions.

examples/example_uq_bootstrap.py
yonatank93: > Let's first make sure we are on the same footing. Below is a general way to do bootstrapping:
> 
> 1. sample training set with replacement
> 2. train the model with the sampled data
> 3. and redo 1 and 2 multiple times
> 4. evaluate on a holdout test set (mean and stdev mostly)

Yes, this is the general way to do bootstrapping.

> You see here we do not have the training on entire training set step. It does not have anything to do with running time, but how we need to do it. My argument was that we don't need to do the training on the entire training set step. It might be ok to do it, but it is then not the typical bootstrapping approach.
> 
> Also, I don't think, in principle, the optimizations in different runs need to use the same optimizing hyerparams, although it is typically the case.

I see what you meant. I will argue that we need to keep the same optimizing hyperparameters. We can think that there is a mapping from the bootstrap dataset to the inferred parameters. For each run, we suggest using the same mapping, i.e., the same optimizing hyperparameters. Otherwise, the mapping from the bootstrap dataset to the parameters is different for each run, and the inferred parameters cannot be compared directly.

In the example, I trained the potential before running bootstrapping to set the optimizing hyperparameters to use in the bootstrapping. I guess, this step (training to the original dataset) is not necessary, but using the same optimizer setting is an important step in bootstrapping.

yonatank93:
@mjwen I applied what we just discussed. Now I see that adding the argument `seed` is more elegant, like what you suggested. I can also see how this automatically leads to reproducible UQ results, even if the user doesn't pay attention to it. I use the seed to create a local random number generator, which I think satisfies some of my reasoning above too.

I also ran pre-commit and built the documentation, but you might want to double-check it. It is ready for you to review again.

mjwen:
We have CI checking the formatting, using the same pre-commit configs. So as long as it passes the checks, it is good. 

Everything is good now. I've merged it. Thanks!

yonatank93:
Thanks for your help.

PR #119, opened 2023-04-18
Updated functionality to load from ColabFit

EFuem:
Fixed how properties are handled within the new data standard of ColabFit. Tested and works for energy, forces, and stress.


mjwen:
@ipcamit can you please review this? For some reason that I don't know, I cannot request you to be a reviewer using the GH interface.

ipcamit:
Ok. Let me check once

mjwen:
@EFuem @ipcamit Thanks both! the test is not passing, because of kim-api issue in the CI. I've fixed it in the main branch, but not in the ml branch. I will merge this and ensure the test passes when we merge all the new ml stuff to master.

PR #132, opened 2023-09-21
First changes for `Parameter` class

ipcamit:
## Summary

First major commit toward decoupling `Parameters` from `models`
* All of parameter related functions are localized in parameter class
* tests now reflect newer interface

## Deleted tests
* The UQ tests were removed as it will take some more time to made adequate changes

## TODO
There is certain level of duplication in functionality. Specially between `set_opt_param` and `set_param_mutable`. `set_param_mutable` is supposed to mark the parameters which can mutate during training, while their values can be set in models using `set_opt_param`. 

Originally `set_opt_param` did both of these tasks so it was fine. This is little confusing now as I guess in current framework more appropriate naming should be `set_param_value` etc. which would clash with update model params function.

So would try to comeup with more coherent API in future. Hence also holding on modifying UQ stuff till this interface is more mature.

## Next up
* ~~dataset class: mainly colabfit integration that Eric provided and capabilities to read multifile xyz~~
* Property Transformations


mjwen:
I guess this is not ready for review yet, right? If ready, ping me and I will take a look.

ipcamit:
Sure. I was holding out for libdescriptor as it would be used in next transformation module of kliff. Now that I have SOAP implemented there, I will write to you and yonathan again for discussing design of new `transformation` module to contain all possible transforms. Afterwards we can start merging the changes

ipcamit:
Will raise new pull request with just dataset class

PR #137, opened 2023-11-15
Update the ptemcee dependency

yonatank93:
The original ptemcee repo is not maintained. I changed the dependency to be my fork of the ptemcee repo. I have a minimal fix in my fork compared to the original repo, but it fixes some incompatibility, especially with newer numpy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,14 @@ setup(
     # This also prohibits the use of py3.10; most distribution channels (e.g. conda)
     # do not have numpy<1.20 support anymore.
     extras_require={
-        "test": ["pytest", "kimpy", "ptemcee", "emcee", "torch", "numpy<1.20"],
+        "test": [
+            "pytest",
+            "kimpy",
+            "emcee",
+            "git+https://github.com/yonatank93/ptemcee.git@enhance_v1.0.0",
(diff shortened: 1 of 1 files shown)


yonatank93:
@mjwen Just created a PR for the update for ptemcee dependency.

mjwen:
@yonatank93  The CI test is failing. I took a look and seems there might be some specific syntax to add a git repo to `extra_requires`. 

Can you try installing it locally and see whether we still have the error as in the CI test, and if yes, we need to fix it.

yonatank93:
@mjwen I was able to reproduce the issue. I think the problem is how I listed ptemcee in `extra_requires`. I think I found a way to fix it, and let's see if it works.

mjwen:
@yonatank93 Yeah, it worked. Thanks! 

One more request, can you delete the comments above `extra_requires`? Then we can merge it.

yonatank93:
@mwen I just did it. Let me know if there are any changes that you want me to make. I also haven't worked on the uq test, as you suggested.

mjwen:
Let's wait until most of Amit's stuff gets merged and then we can think once more about how to adapt the UQ tests.

PR #153, opened 2024-01-08
Added complete transform module for graph generation and 

ipcamit:
## Summary
This PR integrates all the transforms for the Configuration objects, including configuration transforms (fingerprinting), and property transforms (normalization of energy and forces).
The coordinate transforms are the major contribution which includes graph generation and descriptor generation from `libdescriptor`.
It also introduces functions to compute some descriptor widths in `kliff.utils` as I was not sure of where to best keep those functions.
I will commit the unit tests for these modules next, which will install the libdescriptor python package to compute and compare descriptors. That commit will also move existing descriptor module to legacy. But will do all of that in next commit with first round of comments.

Also, with this commit I have modified the README file to state the purpose of this branch, as Stefano asked it for pointing it to people. Please let me know if we should include more information.

PS: @mjwen I am also tagging Eric (@EFuem ) in this PR as he has offered to be an extra set of eyes for helping out in the transition.


kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,469 @@
...
+        self.hyperparameters = self._set_hyperparams(hyperparameters)
+        self.cutoff_function = cutoff_function
+        self._cdesc, self.width = self._init_descriptor_from_kind()
+        if nl_ctx:
+            self.nl_ctx = nl_ctx
+            self.external_nl_ctx = True
+        else:
+            self.external_nl_ctx = False
+
+    def _set_hyperparams(self, hyperparameters):
```
mjwen, line 116: Rename -> `_get_hyperparams`, and make it a `staticmethod`?

kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,469 @@
...
+        self.species = species
+        _available_descriptors = AvailableDescriptors()
+        self.descriptor_name = descriptor
+        self.descriptor_kind = getattr(_available_descriptors, descriptor)
+        self.width = -1
+        self.hyperparameters = self._set_hyperparams(hyperparameters)
+        self.cutoff_function = cutoff_function
+        self._cdesc, self.width = self._init_descriptor_from_kind()
+        if nl_ctx:
+            self.nl_ctx = nl_ctx
```
mjwen, line 111: In the forward and backward functions below, a neighbor list should be created for each config. Here, if a user provides `nl_ctx`, can it be used later in the two functions?

kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,469 @@
...
+
+        TODO:
+            Refactor this function to abstract ut all respective functions. That should
+            make this function easier to maintain. e.g. See _init_descriptor_from_kind().
+
+        Args:
+            path (str): Path to the directory where the file will be saved.
+            fname (str): Name of the descriptor file.
+        """
+        with open(os.path.join(path, fname), "w") as fout:
```
mjwen, line 322: Let's switch to use `pathlib.Path`, not `os.path`

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+            return 0
+
+    def __cat_dim__(self, key, value, *args, **kwargs):
+        if "index" in key or "face" in key:
+            return 1
+        else:
+            return 0
+
+
+class KLIFFTorchGraphGenerator(ConfigurationTransform):
```
mjwen, line 62: Can we rename it to something like `KIMDriverGraph`?

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+            self.cutoff,
+            configuration.species,
+            configuration.coords,
+            configuration.cell,
+            configuration.PBC,
+        )
+        graph.energy = configuration.energy
+        graph.forces = configuration.forces
+        if torch_available:
+            if self.as_torch_geometric_data:
```
mjwen, line 119: Combine these two if checks ?

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+            self.n_layers,
+            self.cutoff,
+            configuration.species,
+            configuration.coords,
+            configuration.cell,
+            configuration.PBC,
+        )
+        graph.energy = configuration.energy
+        graph.forces = configuration.forces
+        if torch_available:
```
mjwen, line 118: This `if torch_available` is error prone, for example, in the case a user wants to export to PyG graph, but forget to install torch.

Essentially, this should only be detemined by `self.as_torch_geometric_data`. If it is true, we should export a PyG graph. We can raise error if torch is not available.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+                f"edge_index{i}", torch.as_tensor(graph.edge_index[i])
+            )
+        # torch_geom_graph.coords.requires_grad_(True)
+        return torch_geom_graph
+
+    def collate_fn_single_conf(self, config_list):
+        """
+        Collate function for use with a Pytorch DataLoader. This function is used when
+        dataloader loads each configuration as a batch of size 1. This is useful when
+        the model is trained on a single configuration at a time.
```
mjwen, line 155: Can we provide a general `collate_fn`? 

Although the current model might be trained using a single config, but in general we want to support any batch size.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+        implicit_fingerprint_copying (bool): If True, the fingerprint will be copied to
+            the Configuration object's fingerprint attribute.
+    """
+
+    def __init__(
+        self,
+        species,
+        cutoff,
+        n_layers,
+        as_torch_geometric_data=False,
```
mjwen, line 84: I've remove `as_torch_geometric_data` here, and `self._to_py_graph(graph)` in __init__. Instead, changing `self._to_py_graph(graph)` to `self.to_pyg_graph(graph)` and call it explicitly when necessary. 

This should make the class more clear.

mjwen (changes requested):

mjwen:
> PS: @mjwen I am also tagging Eric (@EFuem ) in this PR as he has offered to be an extra set of eyes for helping out in the transition.

This is great. Thanks @EFuem !

mjwen:
@ipcamit thanks! 

I don't think there will be major changes needed but a couple of small tweaks. I have no idea of how `KLIFFTorchGraph` is going to be used, and thus weren't be able to provide feedback. 

Also, let's be consist in naming and such:
- let's use `pyg` or `PyG` in function/class names for torch_geometric related stuff?
- let's use `KIMDriver` for the graphs and related stuff for your driver?
- let's remove conditional import as much as possible, which typically can cause a lot of confusion 
- type annotation is needed, but I suppose you plan to do it once everything else is finalized

kliff/transforms/configuration_transforms/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from .configuration_transform import ConfigurationTransform
+from .descriptors import Descriptor, show_available_descriptors
+from .graphs import *
+
+if torch_geometric_available():
+    __all__ = [
+        "ConfigurationTransform",
+        "Descriptor",
+        "KLIFFTorchGraphGenerator",
+        "KLIFFTorchGraph",
```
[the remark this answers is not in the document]
ipcamit: * Removed conditional imports. 
* dummy graph module is imported if PyG is not found

kliff/transforms/configuration_transforms/descriptors.py
ipcamit: This comment is unclear, isn't that what is happening? if user provides external neigbor list context then it is used in forward and backward function. Otherwise it will compute own neigbor lists. But I think it might not be that good of a design. It was supposed to work with other neigbor lists and reusing precomputed neighborlists. With implicit fingerprint copying I think I can remove this one.

ipcamit: Does that answer your question?

kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,469 @@
...
+                return symmetry_functions_set51()
+            elif hyperparameters == "set30":
+                return symmetry_functions_set30()
+            elif hyperparameters == "bs_defaults":
+                return bispectrum_default()
+            elif hyperparameters == "soap_defaults":
+                return soap_default()
+            else:
+                raise DescriptorsError("Hyperparameter set not found")
+        elif isinstance(hyperparameters, OrderedDict):
```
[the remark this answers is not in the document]
ipcamit: switched from OrderedDict to Dict.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    def __init__(self):
+        super(KLIFFTorchGraph, self).__init__()
+        self.num_nodes = (
+            None  # Simplify sizes and frees up pos key word, coords is cleaner
+        )
+        self.energy = None
+        self.forces = None
+        self.n_layers = None
```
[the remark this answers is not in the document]
ipcamit: The major issue that arises in converting GNN to KIM-API is the graph updates. Most, if not all, GNNs uses a flavor of minimum image convention kind of node attribute list, where you only need graphs of atoms upto cutoff distance. In KIM-API we do not et global lattice information to do so. Hence graphs in KIM-API has to be reconstructed from purely cutoff/influence distance information. Therefore what this class do is that it calculates these graphs as convolution stages. That it if you want a 3 layer conv graph, it will generate graphs such that `infl_dist = cutoff * n_layers` and you get `edge_index0`, `edge_index1`, and `edge_index2` graphs where you can now apply the convolution operations in sequential fashion
```
h = conv(edge_index0, h)
h = conv(edge_index1, h)
h = conv(edge_index2, h)
```
as opposed to more conventional
```
for i in range(3):
   h = conv(edge_index,h)
```

This does have a benefit of making these operations parallel, as you no longer need global information. But they are essential for making graphs compatible to KIM-API.
Hence This class aggregates all the information needed for the graphs, including `n_layers`

Hope this helps.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: Done. `to_py_graph` now returns `PyGGraph` class, which is initialized as a dummy class with same attributes if PyG is not available.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: See above. Now `PyGGraph` is always used. If PyG is present then it will be an instance of `pyg.data.Data` otherwise it will be a plain Python class with same fields. In the latter case it will log warning as well.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: Removed both

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: I am leaving this comment up for now. Because I need to investigate a bit more. A general `collate_fn` is part of base class, so that might not be needed, if I remember correctly this was needed for making the PyG collation properly. But I think we can remove it now. Perhaps in next commit.

kliff/transforms/configuration_transforms/graphs/graph_generator_python_bindings.cpp
```diff
@@ -0,0 +1,297 @@
...
+ * is copied here. This is done to avoid the need complicated build steps
+ * and to make the graph module self sufficient.
+
+ * Author: Amit Gupta
+ * Email: gupta839@umn.edu
+
+ * License: LGPL v2.1, This is distributed under the original KLIFF license.
+ */
+
+#include <cstddef>
```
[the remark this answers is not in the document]
ipcamit: renamed to `kim_driver_graph_data.cpp`

kliff/transforms/configuration_transforms/graphs/helper.hpp
```diff
@@ -0,0 +1,116 @@
+// Author: Mingjian Wen (wenxx151@gmail.com)
```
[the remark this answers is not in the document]
ipcamit: No need of symlinking, I had another solution in case this becomes untenable. I can just use setup.py to properly link to original files. Let me know if that works. Updated the setup.py

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Attributes:
+        keep_original: If True, the original property values are kept in the configuration
+            object. Otherwise, the original property values are discarded.
+        original_property_value_map: A map that stores the original property values of
+            configurations. It stores list of property values for each configuration.
+        property_key: The key of the property to be transformed.
+    """
+
+    def __int__(self, property_key="energy", keep_original=True):
```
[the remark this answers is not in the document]
ipcamit: huh, surprising the static analyzer didnt catch it.

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,189 @@
...
+class PropertyTransform:
+    """
+    A property transform is a function that maps a property of a configuration to a
+    transformed property. For example, current property transforms include normalization
+    of the energy and forces.
+
+    Attributes:
+        keep_original: If True, the original property values are kept in the configuration
+            object. Otherwise, the original property values are discarded.
+        original_property_value_map: A map that stores the original property values of
```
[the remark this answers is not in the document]
ipcamit: I have no issues with it. But that would mean all the property transforms need to be bijective. The main example I had in mind was ML models to predict norm of forces, so that scalar values can be predicted per atom-pair, which can then be used to generate force pairs per vector. Leaving this unresolved for now. Will remove if needed in next iteration.

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,189 @@
...
+        self.property_key = "energy"
+
+    def transform(self, dataset: Union[List["Configuration"], "Dataset"]):
+        """
+        Transform the property of a configuration.
+
+        Args:
+            dataset: A list of configurations or a dataset.
+        """
+        raise PropertyTransformError("This method is not implemented.")
```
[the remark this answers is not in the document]
ipcamit: I was matching the existing style. I actually liked the idea of per module errors, as they are really clear about where the error occured. Keeping it for now, and we can remove it all together if needed.

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,189 @@
...
+    def transform(self, dataset: Union[List["Configuration"], "Dataset"]):
+        configuration_list = self.get_configuration_list(dataset)
+        n_configs = len(configuration_list)
+        original_property_values = list(
+            map(lambda config: getattr(config, self.property_key), configuration_list)
+        )
+        original_property_values = np.vstack(original_property_values)
+        if self.keep_original:
+            self.original_property_value_map = original_property_values
+        self.mean = original_property_values.mean()
```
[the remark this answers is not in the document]
ipcamit: Do you mean as an example? I mean why add one more dependency?

kliff/utils.py
```diff
@@ -213,3 +222,42 @@ def stress_to_tensor(input_stress: list) -> np.ndarray:
     stress[0, 1] = stress[1, 0] = input_stress[5]
 
     return stress
+
+
+def get_bs_size(twojmax, diagonal):
```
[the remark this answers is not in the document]
ipcamit: moved to `configguration_transforms/utils`

ipcamit (commented):
I think I have addressed the comments, please let me know if you have more suggestions. Pushing the changes after this. Just saw comment about type annotations. Will include it in next iteration.

kliff/transforms/configuration_transforms/graphs/helper.hpp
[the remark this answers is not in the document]
mjwen: That is even better!

kliff/transforms/property_transforms.py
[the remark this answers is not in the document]
mjwen: it just ignores this method

kliff/transforms/property_transforms.py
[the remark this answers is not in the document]
mjwen: sounds good, we can discuss this later

kliff/transforms/property_transforms.py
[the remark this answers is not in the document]
mjwen: No, I meant to replace the manual implementation. 

I agree with you that this class is relatively simple and easy to implement ; no need to use sklearn

kliff/transforms/property_transforms.py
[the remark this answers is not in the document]
mjwen: right, I was only thinking of the bijective case. Then we should keep it.

kliff/transforms/configuration_transforms/graphs/graph_generator_python_bindings.cpp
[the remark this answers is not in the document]
mjwen: I meant rename `the graph_generator_phthon_bindings.cpp` to one of the two provided names. All good now.

kliff/transforms/configuration_transforms/configuration_transform.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Returns:
+            Fingerprint of the configuration.
+        """
+        raise NotImplementedError
+
+    def __call__(self, configuration: "Configuration"):
+        fingerprint = self.forward(configuration)
+        if self.implicit_fingerprint_copying:
+            configuration.fingerprint(fingerprint)
```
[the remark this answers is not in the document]
mjwen: We might need to modify the old KLIFF. 

I meant if `implicit_fingerprint_copying` is true, we attach it to the corresponding configuration but return the fingerprint whatsoever.

mjwen:
@ipcamit thanks! 

I'll give it a second go soon and provide feedback if I have.

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,197 @@
...
+            property_to_transform = getattr(configuration, self.property_key)
+            property_to_transform *= self.rms_mean
+            setattr(configuration, self.property_key, property_to_transform)
+
+
+class RMSMagnitudeNormalizePropertyTransform(PropertyTransform):
+    """
+    Normalize the property of a configuration to zero mean and unit variance,
+    using the root mean square of the magnitude of the property. This method is
+    useful for normalizing forces.
```
mjwen, line 158: More docs need to explain this and `RMSNormalizePropertyTransform`. Ideally, providing the mathematical formula on how it is computed?

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, property_key="energy", keep_original=True):
+        super().__init__(property_key, keep_original)
+        self.original_property_value_map = None
+        self.property_key = property_key
+        self._keep_original = keep_original
+        self.mean = 0.0
+        self.std = 1.0
+
+    def transform(self, dataset: Union[List["Configuration"], "Dataset"]):
```
mjwen, line 91: To make it more modular, I propose the below:
```python 
def transform(self, dataset: Union[List["Configuration"], "Dataset"]):
    original_values = self._get_property_values(dataset, keep_original)
    transfromed_values = ... # performed the transformation 
    self._set_property_values(dataset, transformed_values)
```
We just need to implement two more functions `_get_property_values` and `_set_property_values` in the base class. It can be shared and reused by all subclasses.

kliff/transforms/configuration_transforms/graphs/kim_driver_graph_data.cpp
```diff
@@ -0,0 +1,297 @@
+/*
+ * Contains functions to generate complete graph from a given set of
+ * coordinates and species. The generated graph uses the neighbor list
+ * code from the kliff.nl module, but instead of linking against it, it
+ * is copied here. This is done to avoid the need complicated build steps
```
mjwen, line 5: It is not copied any more. Need update.

kliff/transforms/configuration_transforms/graphs/kim_driver_graph_data.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include <iostream>
+#include <string>
+#include <vector>
+#include <set>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+#include <pybind11/stl.h>
+
+#include "neighbor_list.h"
```
mjwen, line 25: Although you've added something in `setup.py`, but here we are still directly accessing it, without providing any path. Not sure whether this works or not.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,218 @@
...
+    class PyGGraph:
+        """
+        A dummy class to when torch geometric is not available. It has same attributes
+        as PyGGraph class, but does not inherit from torch_geometric.data.Data.
+        """
+
+        warning_logged = False
+
+        def __init__(self, *args, **kwargs):
+            if not PyGGraph.warning_logged:
```
mjwen, line 73: Currently, it will always give a warning when instantiating it. Let's remove this check and `warning_logged`?

kliff/transforms/configuration_transforms/descriptors.py
mjwen: I leave a comment. See that.

kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,477 @@
...
+        self.cutoff = cutoff
+        self.species = species
+        _available_descriptors = AvailableDescriptors()
+        self.descriptor_name = descriptor
+        self.descriptor_kind = getattr(_available_descriptors, descriptor)
+        self.width = -1
+        self.hyperparameters = self.get_default_hyperparams(hyperparameters)
+        self.cutoff_function = cutoff_function
+        self._cdesc, self.width = self._init_descriptor_from_kind()
+        if nl_ctx:
```
mjwen, line 115: `Descriptor` is a subclass of `ConfigurationTransform`, so I expect `forward` and `backward` can be called multiple times, e.g. each time for a different configuration. 

However, `NeighborList` is bound to configuration, i.e. a `NeighborList` is needed for each config. 

So, I don't see how a single `nl_ctx=NeighborList` passed in __init__ can work for different configurations in `forward` and `backward`. It might be possible if one provides something else other than `NeighborList` to `nl_ctx`. Do we have that already? If not, I'd suggest removing `nl_ctx` to simplify things. If we want them, we can add back later.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,218 @@
...
+
+        def __cat_dim__(self, key, value, *args, **kwargs):
+            if "index" in key or "face" in key:
+                return 1
+            else:
+                return 0
+
+else:
+
+    class PyGGraph:
```
mjwen, line 64: What is the use case of this class? 

It will not work with PyG, nor with KIMDriver.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,218 @@
...
+        copy_to_config (bool): If True, the fingerprint will be copied to
+            the Configuration object's fingerprint attribute.
+    """
+
+    def __init__(
+        self,
+        species,
+        cutoff,
+        n_layers,
+        copy_to_config=False,
```
mjwen, line 110: This is not descriptor and I believe there is nothing to copy. So, remove it in __init__, and set it to `False` in super()?

Update:
Well, we can attach a graph to a config in principle. I'm OK to keep this. But please update the docstring above.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        Convert a C++ graph object to a KLIFF Geometric Graph Data object, ``GraphData``.
+
+        Args:
+            graph: C++ graph object.
+
+        Returns:
+            PyGGraph object.
+        """
+        torch_geom_graph = PyGGraph()
```
mjwen, line 153: Very minor: torch_geom_graph -> pyg_graph?

mjwen (changes requested):

mjwen:
@ipcamit A few more comments. The only major one is on `NeighborList`. We can discuss further.

ipcamit:
sure. let me go through these and then we can discuss further.

kliff/transforms/configuration_transforms/configuration_transform.py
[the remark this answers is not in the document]
ipcamit: Great, so exactly like the modified version:
```python
        if self.copy_to_config:
            configuration.fingerprint(fingerprint)
        return fingerprint
```

kliff/transforms/configuration_transforms/descriptors.py
ipcamit: Ok. I think I know from where this bug crept up. It was orginally written as an torch.Autograd function thus the forward and backward were static function with independent `ctx` context being passed around. I modified it incorrectly to its current form. 

Edited it now to not have neighbor context.

kliff/transforms/configuration_transforms/descriptors.py
ipcamit: Replied to previous comment.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: What do mean it will not work with PyG? It is direct child of `pyg.data.Data` class, it is supposed to work with PyG dataloaders, such that it will collate the graphs into a single graph.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: Sorry I dont understand. Without the `warning_logged` it will warn the user n times for instantiation. With the warning logged it is only logging the warning once, try the example below:
```python
from loguru import logger

# log once
class PyGGraph:
        warning_logged = False

        def __init__(self, *args, **kwargs):
            if not PyGGraph.warning_logged:
                logger.warning("Pytorch Geometric not available, using dummy PyGGraph.")
                PyGGraph.warning_logged = True

# log one time
print("Instantiation with class variable warning_logged")
graphs = []
for i in range(5):
    graphs.append(PyGGraph())


# log every time
print("Instantiation without class variable warning_logged")
class PyGGraph:
        def __init__(self, *args, **kwargs):
            logger.warning("Pytorch Geometric not available, using dummy PyGGraph.")

graphs = []
for i in range(5):
    graphs.append(PyGGraph())
```

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: Can you clarify which portion of docstring you think needs changing? Updated fingerprint -> graph.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: OK. Removed for now. If needed we can think of it later (when integrating unit tests).

setup.py
```diff
@@ -71,6 +73,21 @@ def get_extra_compile_args():
     language="c++",
 )
 
+graph_module = Extension(
```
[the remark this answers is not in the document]
ipcamit: I have included `graph_module` to `ext_modules`. Modified `get_includes` function to also append `kliff/neighbor` as I guess thats the appropriate function for it? I can write a separate function too if this was not the intended use of `ext_modules`. KLIFF can now compile and import `graph_module` on my laptop.

kliff/transforms/configuration_transforms/graphs/kim_driver_graph_data.cpp
ipcamit: In the latest setup.py it works.

ipcamit (commented):
Addressed all comments. I think it is nearly done. Will push updated code.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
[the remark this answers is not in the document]
ipcamit: I would need to think a bit more on this to comment on if they are subset or not. I think they are, but need to confirm.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
mjwen: The one under `if`, i.e. `class PyGGraph(Data)` makes sense. The one after `else:` is not derived from pyg Data. There is no collate_fn as well. So, I'd expect it will not work. That's why I was asking the question.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
mjwen: I was not talking about docs, but the real code, `copy_to_config=False`. I think the current setting is good. In principle, one can attach a graph to a config.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
mjwen: I understand now. I misunderstood `PyGGraph.warning_logged = True` as `self.warning_logged = True`. Good now.

mjwen (commented):
All good. The only question I have now is the `PyGgraph`. I commented in the original thread. Copied here: 

```python
else:

    class PyGGraph:
```

The above defined is not a subclass of pyg data. So, it will not have collate_fn and such. Then what's its use case?

kliff/transforms/configuration_transforms/graphs/generate_graph.py
mjwen: I understand now. I misunderstood `PyGGraph.warning_logged = True` as `self.warning_logged = True`. Good now.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: So are there no more changes pending here? Shall I mark it resolved?

mjwen: yes. Marked it.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
ipcamit: The `else` PyGData class is just a container for the data that can still work as in graph models. It is just that it will not collate directly. Hence user will get warning once. Other option would include conditional import of classes which we are slowly moving away from. Calling collate function on this should result in error as it will not have requisite dunder functions. But for single config cases, this or the torch geometric graph will work identically.

ipcamit:
I have commented in detail above. The main use case is to avoid any conditional imports of the dataclass. Without torch geometric it will log a warning and work as a simple data container. With torch geometric it will also collate. So with or without torch geometric KLIFF code will work. With Torch geometric it will be just more efficient.

kliff/transforms/configuration_transforms/graphs/generate_graph.py
mjwen: I'd suggest we remove this. First, two classes with the same name but different behaviors typically is a source of confusion and error. Second, we can add PyG as a dependence. It is just a python package and easy to install.

mjwen: PyG is available from the conda-forge channel. https://github.com/conda-forge/pytorch_geometric-feedstock

ipcamit: OK. Would remove it and submit new commit.

ipcamit: So KLIFF will have Pytorch geometric as hard dependency?

mjwen: Probably not. 

My main concern is that people who are only interested in physics-based models won't need all these. So, ideally, PyTorch and PyG should only be optional. We just issue run time errors using `@requires` and give detailed installation guides. I think that would be it. Honestly, I don't think there are many people in this category.

ipcamit: Sure, no issues. Would remove the `else` `PyGGraph` defination and use `@requires`

mjwen:
All good. I think the only missing part is type annotation. Can you please add that?

ipcamit:
I have added type annotations wherever I could. Please let me know if you spot any missed ones.

mjwen (changes requested):
- A couple of annotation needs to be added in `__init__`
- `nl_ctx` in one methods need to be removed

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+class NormalizedPropertyTransform(PropertyTransform):
+    """
+    Normalize the property of a configuration to zero mean and unit variance.
+    .. math::
+        x' = \\frac{x - \\mu}{\\sigma}
+    """
+
+    def __init__(self, property_key="energy", keep_original=False):
```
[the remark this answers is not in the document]
ipcamit: Do kliff add annotations to children classes as well? These functions were left as they do not modify the parent class functions. I can add them, but I thought kliff did not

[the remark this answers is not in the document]
mjwen: Yes, please. I think we should.

kliff/transforms/configuration_transforms/descriptors.py
```diff
@@ -0,0 +1,470 @@
...
+    """
+
+    def __init__(
+        self,
+        cutoff: float,
+        species: List[str],
+        descriptor: str,
+        hyperparameters: Union[Dict, str],
+        cutoff_function: str = "cos",
+        nl_ctx: NeighborList = None,
```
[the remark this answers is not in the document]
ipcamit: Yes. It was scrubbed from the code, accidentally left in function argument. Removed from docstring and there.

kliff/transforms/property_transforms.py
```diff
@@ -0,0 +1,210 @@
...
+        self._keep_original = keep_original
+        self.original_values = None
+        self.property_key = property_key
+
+    @property
+    def keep_original(self):
+        return self._keep_original
+
+    def transform(self, dataset: Union[List["Configuration"], "Dataset"]):
```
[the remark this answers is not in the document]
ipcamit: I guess only parameter transforms will cause circular imports. I tried removing them from everywhere but parameter transforms and the it imports fine. So removed them from configuration and property transforms.

PR #156, opened 2024-01-29
Fix tests due to function name changes 

mjwen:


mjwen:
@ipcamit I've fixed the tests and merged into v1. 

For the next PR you have, we should make all tests pass, unless some major breaking changes are preventing us from doing that.

ipcamit:
No. With the last merge I think I can get all the tests up until the transformation modules pass. I am currently adding some functionality to libdescriptor, will then add tests to kliff for all the changes made till now.

PR #161, opened 2024-02-06
Order of arguments path and ase_atoms_list was switched in Dataset

marcoscaa:
## Summary

Two minor changes in dataset.py. Order of arguments given to add_from_ase and _read_from_ase was switched in two cases:

* Dataset.from_ase: instance.add_from_ase(...)
* Dataset.add_from_ase: self._read_from_ase(...)


mjwen:
Thanks @marcoscaa!

PR #162, opened 2024-02-08
Colabfit enhancments by Eric

ipcamit:
This PR consists changes made by Eric to colabfit io. List of changes:

Edited functions:
1. `Dataset.from_colabfit`
2. `Dataset.add_from_colabfit`
3. `Dataset._read_from_coabfit`
4. `Configuration.from_colabfit`

Added functions:
1. `Configuration.to_ase_atoms`: converts KLIFF configuration to ASE atoms configuration. Used for storing the KLIFF dataset back to Colabfit.

Other changes:
using from `chemical_symbols` `ase.data`

### DIFF
--- a/kliff/dataset/dataset.py
+++ b/kliff/dataset/dataset.py
@@ -10,7 +10,7 @@ from monty.dev import requires
 
 from kliff.dataset.extxyz import read_extxyz, write_extxyz
 from kliff.dataset.weight import Weight
-from kliff.utils import stress_to_voigt, to_path
+from kliff.utils import stress_to_voigt, stress_to_tensor, to_path
 
 # For type checking
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


mjwen (changes requested):
the order of `ase_atoms_list`, and `path` should be switched in 
```python
instance.add_from_ase(
            ase_atoms_list, path, weight, energy_key, forces_key, slices, file_format
        )
        return instance
```
as well. 

It is called in `from_ase` method. See #161

mjwen:
Are we planning to add tests for this class, so that we can avoid such bugs?

ipcamit:
Yes. Next PR would be of tests for this and transforms modules. It was on hold has I had to fix the descriptor library first. Finished it this week, so KLIFF tests are next.

PR #164, opened 2024-02-20
Added tests for newer modules

ipcamit:
+ Fixed but in voigt stress conversion functions
+ Fixed implicit fingerprinting bug in configuration transforms
+ Fixed documentation and typing inaccuracies
+ Added tests for ASE dataset/configuration interactions
+ Added tests utils (stress functions currently)
+ Added tests for configuration transformations
   + libdescriptor integrations and comparison with older modules
   + graph generations
   + Abstract `ConfigurationTransform` class
+ Added tests for `PropertyTransform`

I think this covers test cases for most of the newer contributions. Let me know if I missed something. I would start working on the trainer modules now.


tests/transformations/test_configuration_transform.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+from ase.data import atomic_numbers
+
+from kliff.dataset import Configuration
+from kliff.transforms.configuration_transforms import ConfigurationTransform
+
+
+class GlobalCoulombMatrix(ConfigurationTransform):
```
mjwen, line 8: Any reason why creating a new class to test it -- can we use any existing one to test it?

mjwen:
Looks great! Just have a minor question. See above.

tests/transformations/test_configuration_transform.py
ipcamit: I was thinking of something like the LennardJones model test of Model class, to have a trivial implementation of a children class for checking any inconsistency. Although LennardJones is a member of `models` module, I wan not sure if Coulomb matrix will be of much use. Either I can move it in main code, or remove it all together, as anyway Descriptor and graph transforms are being tested.

mjwen: This is ConfigurationTransform -- how is LennardJones (which is a model) related?

ipcamit: It is not related. I meant similar design principle. Like how LJ extended Model class for a simple system to test all the Parent class routines, I was thinking this globalCM class can be used to test routines of parent ConfigurationTransform class.

mjwen: I see. 

We can then probably leave the CM stuff in the test folder.

PR #172, opened 2024-03-04
Design Review 1: KLIFF Trainer framework

ipcamit:
Idea: A coherent fully reproducible framework to train all supported openkim models (including ML ones) and be able to archive exact training methodology for reuse and be available from openkim.

So easiest is to create a yaml file which can be hashed to ensure integrity and this hash can be included in `kimspec.edn` etc for provenance. This PR request contains first draft of this framework, along with its implementation for KIM physics based models. I think better to have a look at it now before having to review huge changes at once. Example yaml file is also there at the bottom.

Core changes and contributions:
1. `Dataset`: dataset object can now load weights from a file (I think this was also there on one of the todo list). the weights file has to be a 4 column whitespaced ascii file, that numpy can directly load. The 4 columns represent the configuration, energy, forces and stress weights. The length of the file has to be either 1 (all weights are same) or n (where n is the number of configurations the weights are to be set). In both ml and kim based training, these per configuration weights will be used.
2. `_exceptions`: This is just a proposal, and not very strong one at that, but I think it might be useful to collect all of the file based exceptions, and collect them under single file. This gives the flexibility to use them anywhere in the KLIFF more freely. Currently we might run in circular imports if two related modules want to use same exception. (Example Trainer base class and KIMtrainer should raise Trainer error). Let me know your thoughts.
3. `kim_residuals`: simple collection of most used loss functions. Another idea is to save the pickled version of the loss function for reproducibility. 
4. `kliff_trainer`: This is the base class for trainer framework. All trainers must inherit this class. This includes some basic functionality:
    **a.** Initializing the class members from yaml / dict
    **b.** Seed all modules, numpy, scipy, torch etc
    **c.** Setup workspace folder and model folders etc where each independent run would be automatically timestamped and saved
    **d.** setup dataset, load the data, apply any transformation of properties, like energy normalization, compute fingerprint objects, hash the dataset based on path, transforms and weights, save the processed dataset as dill pickle object. Now when the next time any trainer need same dataset configuration, it can be reloaded directly.
    **e.** setup  test train split, either based on indices files, or ranges etc, split the dataset in test data and train data
    **f.** call trainer specific functions, `setup_model`, `setup_parameter_transforms`, `setup_optimizer`
    **g.** hash and archive the training configuration
5. `kim_trainer`: First example of subclassing this trainer class for KIM physics based models. It basically extends `setup_model`, `setup_parameter_transforms`, `setup_optimizer`, and `save_kim_model`
6. enumerations: enumerated list of supported options.

Tar model: KIM model tarred in  a single file. For future proofing against LLNL/KIMKit.

Question: Is there an easy for kimpy to tell me what is the model driver that a model is using.

Future direction: I would like to have a Trainer eventually for each model that can be submitted to the openkim. for v1 we will have Torch and KIM. The Torch trainer will also have a companion Lightning based trainer for v1. 
But for future versions I think it would make sense if we have Trainers that wrap around library like QUIP to provide complete training surface for newer QUIP driver based models and so on. I would also like to have Trainer built around bootstrapping and MCMC sampling that yonathan contributed, but it will take a little more time hence perhaps v1.2 etc. For v1 the UC part will remain unchanged.

Let me know of thoughts and questions. Sorry for huge wall of text, but I think had I delayed it more, it would only have increased in size!

Example yaml file:
<details>

```yaml# Example yaml for kliff trainer
workspace:
    name: test_run # Name of the base workspace folder, where all the runs will be stored
    seed: 12345    # Seed for random number generator, all
    resume: False  # Resume training from a previous run
    walltime: 2:00:00 # Walltime for the run

dataset:
    type: ase           # ase or path or colabfit
    path: Si.xyz        # Path to the dataset
    save: False         # Save processed dataset to a file
    shuffle: False      # Shuffle the dataset
    keys:
        energy: Energy  # Key for energy, if ase dataset is used
        forces: forces  # Key for forces, if ase dataset is used
    training_dataset:
        train_size: 3   # Number of training samples
        train_indices:  # files with indices [optional]
    val_dataset:
        val_size: 1     # Number of validation samples
        val_indices:    # files with indices [optional]
    test_dataset:
        test_size:
        test_indices:
    colabfit_dataset:
        dataset_name:
        database_name:
        database_url:

model:
    model_type: kim     # kim or torch or tar
    model_path: ./
    model_name: SW_StillingerWeber_1985_Si__MO_405512056662_006

transforms:
     property:     # optional: path to property transform file
       name: NormalizedPropertyTransform 
       instance:   # optional: instance of custom property transform
       property_key: energy  # optional: key for the property to transform

    parameter:
        instance:
        parameter_list: # optional for KIM models, list of parameters to optimize
          - A:          # doct means the parameter is transformed
            name: LogParameterTransform
            value: 0.1
            bounds: [[-5, 5]]
          - B          # these are the parameters that are not transformed
          - sigma

    configuration: # optional: generate fingerprints from the configuration
        name: Descriptor
        kwargs: 
            cutoff: 3.7
            species: ["Si"]
            descriptor: "SymmetryFunctions"
            hyperparameters: set51
        instance:

training:
    loss:
        function: MSE
        weights: # optional: path to weights file
            energy: 1.0
            forces: 1.0
            config: 1.0
            stress: 0.0
        normalize_per_atom: true
    optimizer:
        name: L-BFGS-B
        provider: scipy
        learning_rate: 
        kwargs:
          tol: 1.e-6 # 1. is necessary, 1e-6 is treated as string

    batch_size: 1
    epochs: 1
    device: cpu
    num_workers: 2
    chkpt_interval: 1
    stop_condition:
    verbose: True

export: # optional: export the trained model
    model_type: tar # kim or tar
    model_path: ./
    model_name: SW_StillingerWeber_trained_1985_Si__MO_405512056662_006

```
<details>


kliff/trainer/kim_residuals.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Dict
+
+import numpy as np
+
+
+def MSE_residuals(
```
mjwen, line 6: We can port residuals at https://github.com/openkim/kliff/blob/038c990cd8980e17a2438146e3beaef09c357529/kliff/loss.py#L36 to here. 

Need to think about whether we want to support normalization and such in the residual function.

kliff/trainer/option_enumerations.py
```diff
@@ -0,0 +1,225 @@
...
+    - COLABFIT: uUses ColabFit dataset exchange instance. Uses
+        ~:class:`~kliff.dataset.Dataset.from_colabfit` method.
+    - KLIFF: Uses KLIFF compatible extxyz files path. Uses
+        ~:class:`~kliff.dataset.Dataset.from_path` method.
+    """
+
+    UNDEFINED = -1
+    ASE = 0
+    COLABFIT = 1
+    KLIFF = 2
```
mjwen, line 84: KLIFF can be confusing here. From a user's perspective, everything is kliff. We need to be more specific, like "EXTXYZ"?

setup.py
```diff
@@ -113,7 +113,7 @@ def get_readme():
     version=get_version(),
     packages=find_packages(),
     ext_modules=[sym_fn, bispectrum, neighlist, graph_module],
-    install_requires=["requests", "scipy", "pyyaml", "monty", "loguru"],
+    install_requires=["requests", "scipy", "pyyaml", "monty", "loguru", "dill"],
```
mjwen, line 116: Where is `dill` needed?

kliff/trainer/kliff_trainer.py
```diff
@@ -0,0 +1,791 @@
+import hashlib
```
mjwen, line 1: Let's rename this file to  `base_trainer.py` or `trainer.py`.

kliff/trainer/kliff_trainer.py
```diff
@@ -0,0 +1,791 @@
...
+from loguru import logger
+
+import kliff.transforms.configuration_transforms
+from kliff._exceptions import TrainerError
+from kliff.dataset import Dataset
+from kliff.transforms.configuration_transforms import ConfigurationTransform
+from kliff.transforms.parameter_transforms import ParameterTransform
+from kliff.transforms.property_transforms import PropertyTransform
+
+from ..dataset.weight import Weight
```
mjwen, line 24: Probably OK to use relative import from the current directory, but let's change to use absolute import for all other cases like this one `..dataset.weight`

kliff/trainer/kliff_trainer.py
```diff
@@ -0,0 +1,791 @@
...
+    implement.
+
+    Args:
+        configuration: configuration dictionary
+    """
+
+    def __init__(self, configuration: dict):
+        # workspace variables
+        self.workspace_name = None  # name of default directory, root
+        self.workspace_name = (
```
mjwen, line 53: `workspace_name` defined twice.

kliff/trainer/kliff_trainer.py
```diff
@@ -0,0 +1,791 @@
...
+        )
+        self.optimizer_kwargs = (
+            None  # kwargs for optimizer to bundle all unknown kwargs
+        )
+        self.learning_rate = None  # learning rate for torch
+
+        self.max_epochs = 10000  # maximum number of epochs
+        self.device = "cpu"
+        self.batch_size = 1
+        self.chkpt_interval = 100
```
mjwen, line 122: Let's use self.ckpt_interval?

kliff/trainer/kim_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+                    if input_params != model_params.name:
+                        raise TrainerError(
+                            f"Parameter name mismatch. Expected {model_params.name}, got {input_params}."
+                        )
+                else:
+                    raise TrainerError(
+                        f"Optimizable parameters must be string or value dict. Got {input_params} instead."
+                    )
+
+    def setup_optimizer(self):
```
mjwen, line 159: Should this be implemented as `get_optimizer()` below?

mjwen (changes requested):

mjwen:
Sorry for the late reply! 

> Question: Is there an easy for kimpy to tell me what is the model driver that a model is using. 

I cannot remember. Better check with Ryan. 

I only have major comments for the trainer. 

1. I don't think the trainer need to deal with dataset and such (like dataset initializing and split). Instead, the train/val/test datasets should be set up outside the trainer, and their instances be passed to the `fit` \ `test` functions of the trainer. This is following the lightning approach and I think it is a good approach. 
2. Related to 1, it might be better to only pass the config dict info related to training to the trainer, other stuff (like model, dataset, transforms) all be passed in as an instance. Basically, I am thinking of a trainer like 
```python
Trainer:
    def __init__(model: Union[KIMModel, TorchModel, TarModel], training_config: dict):
        
    
    def fit(train_loader, val_loader):
         
   
    def test(test_loader):
     
    # other necessary functions such as save kim model  
```
As you can see, this is again modeling after lightning. 

3. There exist good packages to deal with configs that we can take advantage of. We don't need to do all the checking and such in the trainer for all hyperparameters, which would be very difficult to maintain in the long run. Specifically, I am considering [omegaconf](https://omegaconf.readthedocs.io/en/2.3_branch/) and [hydra](https://hydra.cc/). The latter uses the former. The nice feather of hydra is that you can define a base config that is shared by all, and customize it for various specific cases. Also, it is composable, making it much easier to separate the configs for the various parts. schnetpack uses it: https://github.com/atomistic-machine-learning/schnetpack/blob/master/src/schnetpack/configs/train.yaml and I've used it in one of my projects as well https://github.com/mjwen/rxnrep/tree/main/configs, which may give you a feeling of what it look like. hydra also have utility functions e.g. for instantiating class https://hydra.cc/docs/advanced/instantiate_objects/overview/, which we can take advantage of instead of using `importlib`.

ipcamit:
> I only have major comments for the trainer.
> 
>     1. I don't think the trainer need to deal with dataset and such (like dataset initializing and split). Instead, the train/val/test datasets should be set up outside the trainer, and their instances be passed to the `fit` \ `test` functions of the trainer. This is following the lightning approach and I think it is a good approach.

I can look into the hydra library, but for datasets, models etc, do you think it will be better to have different utility functions etc to setup datasets and models? Whole idea is to make sure that using this single yaml file you should be able to get same results in term of dataset splits etc.

mjwen:
Sorry for the confusion... Yes, I'm still imaging a **single** yaml file to config all the components, but the single yaml file can be composed from a couple separate files. If one does not like to use multiple files, all the parts can still be put in a single file. 

Anohter major part is that we probably don't want to pass all the yaml configs to the Trainer class, but only pass instantiated instances to it, so we can simplify the Trainer a bit.

mjwen:
I have mixed feelings with hydra -- it can simplify configs management, but tracking and debugging can be a bit more involved. I am not sure whether we want to use it here. Probably you explore it a bit and we can then decide. But even if we don't use hydra, I think we still need to use OmegaConf to read, merge, write config files, instead of the vanilla yaml module.

ipcamit:
> Yes, I'm still imaging a **single** yaml file to config all the components, but the single yaml file can be composed from a couple separate files. If one does not like to use multiple files, all the parts can still be put in a single file.

This condition is already satisfied to some extent. The configuration is split into 6 independent blocks,  

1. workspace: local folder/cache/other information directory.  
2. dataset: instance of KLIFF dataset
3. model: instance of KIMModel or TorchScript model 
4. transforms: instance of property, parameter, configuration transforms 
5. training: loss function, optimizer, verbose 
6. export: export the trained model, KIM-API model or tarball

I guess I can try splitting it into 3 modules, 
1. Dataset: setup dataset, along with test train split
2. Model: setup KIM, Torch, Tar model etc. It will also consume export module
3. Trainer module: workspace management (base class), training loop (child class), optimizer (child class)

Optimizer can be a separate module but it would be difficult to organize as a separate module owing to scipy's different API. Can make a private optimizer module for torch.

Thoughts, suggestions?

mjwen:
Yes, the previous yaml file does serve the purpose of separating them into different parts. We just need to organize the internal codes in that direction a bit. 

I agree with the three-module approach. Particularly, we want to integrate the optimizer with a trainer for the exact reason you've mentioned.

kliff/trainer/kim_residuals.py
ipcamit: Will return to this later, keeping the comment unresolved for reference.

kliff/trainer/option_enumerations.py
ipcamit: Removed enumerations, they were clunky and quite tedious to maintain. Will directly compare strings for now. "kliff" is renamed to path ("folder" might be better choice, but it is trivial to change)

kliff/trainer/kim_trainer.py
ipcamit: setup optimizer was chosen as it might be more involved, like setting up ema, lr decay etc. I have no preference either way.

ipcamit:
1. Simplified the trainer. 
2. Intialization methods for model and datasets.

Note on glossary, everywhere the configuration dictionary is called "manifest", as configuration was becoming too confusing, example "dataset_from_configuration" creates a confusion on whether it uses the `kliff.Configuration` object. Hence I suggest we use "manifest" as the word of choice for yaml configurations. So KLIFF trainer takes in a training manifest, and splits it accordingly in model, dataset and transform manifest, and passes it to appropriate static functions in Dataset and KIMModel, such that they return require objects.

Regarding Omegaconf, I did integrate it, but have to remove it afterwards as it adds lot of needless complexity with no clear advantage. Biggest issue was that it only support "primitive" dataypes in python, something as basic as numpy array needs custom solutions etc. dict is flexible enough to be a better solution I think.

There might be some rough edges, that will clear as we add more functionality like the ML trainer.

ipcamit:
Disregard for the time being, I need to work a bit more on this before opening the PR again.

PR #174, opened 2024-03-28
ColabFit Changes

EFuem:
Modified the way in which data is fetched from ColabFit. Additional logic needed to be built in, so this was done on the ColabFit side which resulted in a simplification on the KLIFF side.

### DIFF
--- a/kliff/dataset/dataset.py
+++ b/kliff/dataset/dataset.py
@@ -224,8 +224,17 @@ class Configuration:
             weight=weight,
         )
         self.metadata = {
-            "data_object": data_object,
+            "do-id": data_object["colabfit-id"],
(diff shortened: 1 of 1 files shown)


mjwen:
@EFuem Thanks ! 

There is some linting error as can be seen from the CI report above. Can you please do a quick pre-commit run to fix that? See https://kliff.readthedocs.io/en/latest/contributing_guide.html#code-style 

Also, longer term, we need to think about how to test the `from_colabfit` method a bit to make sure everything works. It uses MongoDB and such, which make testing more involved and I don't have idea on how to effectively test that. You've got any idea?

mjwen:
Thanks !

PR #178, opened 2024-05-23
Trainer framework

ipcamit:
I dont expect this PR to be immediately merged. But I think I can use a fresh pair of eyes for the trainer. Now it has three trainers, KIM trainer, DNN trainer for descriptors, and GNN trainer using torch lightining. I think at this point I can really use comments and recommendations.


kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+                    if not kwargs:
+                        raise TrainerError(
+                            "Configuration transform module options not provided."
+                        )
+                    configuration_module = configuration_module(
+                        **kwargs, copy_to_config=True
+                    )
+
+                    for config in self.dataset.configs:
+                        _ = configuration_module(config, return_extended_state=True)
```
mjwen, line 460: Unlike property, here configuration transform is only initialized but the actual transformation does not happen. Where is it performed?

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+        """
+        Set up the optimizer based on the provided information. If the optimizer is not
+        provided, it will be loaded from the optimizer_name. If the optimizer_name is not
+        provided, it will raise an error. If the optimizer_provider is scipy, it will be
+        loaded from the scipy.optimize. If the optimizer_provider is torch, it will be
+        loaded from the torch.optim. Left for the derived classes to implement.
+        """
+        raise TrainerError("setup_optimizer not implemented.")
+
+    def setup_test_train_datasets(self):
```
mjwen, line 509: 1. Name should be updated, like `setup_dataset_split`?
2. We only do train and val below; probably add test?

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+
+        val_indices = self.dataset_sample_manifest.get("val_indices")
+        if val_indices is None:
+            val_indices = np.arange(train_size, train_size + val_size, dtype=int)
+        elif isinstance(val_indices, str):
+            val_indices = np.genfromtxt(val_indices, dtype=int)
+        else:
+            TrainerError("val_indices should be a numpy array or a path to a file.")
+
+        if self.dataset_manifest.get("shuffle", False):
```
mjwen, line 562: I don't think `shuffle` here has any major effect. We will need to shuffle at each training epoch if using mini-batch. Instead, if the entire training set is used to compute the gradients, no shuffle is needed.

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+
+        if train_size + val_size > len(self.dataset):
+            raise TrainerError(
+                "Sum of train, val, and test sizes is greater than the dataset size."
+            )
+
+        # check if indices are provided
+        train_indices = self.dataset_sample_manifest.get("train_indices")
+        if train_indices is None:
+            train_indices = np.arange(train_size, dtype=int)
```
mjwen, line 548: This assigns the first `train_size` data points to the trainset, and the next `val_size` to the val set. For now, we at least need to support random split, using something like
```python
indices = np.random.permutation(trian_size + val_size)
train_indices = indices[:train_size]
val_indices = indices[train_size:]
```

kliff/trainer/kim_trainer.py
```diff
@@ -0,0 +1,181 @@
...
+                f"Optimizer not supported: {self.optimizer_manifest['name']}."
+            )
+        optimizer_lib = importlib.import_module(f"scipy.optimize")
+        self.optimizer = getattr(optimizer_lib, "minimize")
+
+    def loss(self, x):
+        """
+        Compute the loss function for the given parameters. It will compute the loss
+        function. It seems like MPI might be only way to make it parallel as the
+        multiprocessing does not work with the KIM models. KIMPY models are not yet
```
mjwen, line 95: The current version does support parallel via multiprocess. If I remember correctly, it gets around to picking the entire KIM model object. 

But I agree -- let's focus on MPI in the future to avoid complicating it.

kliff/trainer/kim_trainer.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if result.success:
+            logger.info(f"Optimization successful: {result.message}")
+            self.model.update_model_params(result.x)
+        else:
+            logger.error(f"Optimization failed: {result.message}")
+
+    def _get_loss_fn(self):
+        if self.loss_manifest["function"].lower() == "mse":
+            return MSE_residuals
```
mjwen, line 164:
```python
else: 
    Raise ValueError()
```

kliff/trainer/kim_trainer.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def save_kim_model(self):
+        if self.export_manifest["model_type"].lower() == "kim":
```
mjwen, line 167: To simplify it, I don't think we need to distinguish `kim` and `tar`, because it seems the only difference is that the latter is the tarred version of the former. This requires change of some other files as well. 

I think we can generally call this a `kim` model. We have the option to tar it or not here. But when reading the kim model (in another place), we can use a utility function to check what is provided and then proceed accordingly. To be more specific, I can think of three modes: 
1. If a simple model name like `SW_xxxxxx` is provided, this will be regarded as the user wanting to use a model in a kim collection (e.g. user); 
2. If a path is provided like, `./SW_xxxxxx`, then the local model at the corresponding path will be installed and used;
3. If a tar path is provided like `./SW_xxxxxx.tgz`, we untar and then install. 

For the three models, we can internally use a utility function check whether it is a name, a path to a directory, or a path to a tarball.

kliff/trainer/torch_trainer.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+        else:
+            return optimizer(
+                self.model.parameters(), lr=self.optimizer_manifest["learning_rate"]
+            )
+        # TODO: Scheduler and ema
+
+    def _get_loss_function(self):
+        if (
+            self.loss_manifest["function"].lower() == "mseloss"
```
mjwen, line 88: We can simplify to force the use of only one, like `mse`.

kliff/trainer/torch_trainer.py
```diff
@@ -0,0 +1,366 @@
+import os
```
mjwen, line 1: Any major reason we want to implement the `torch_trainer.py`? Can it be merged with the lightning trainer to simplify things?

mjwen (commented):
Looks great! Two major comments: 

1. Can we merge `torch_trainer.py` and `lightning_trainer.py`, to use lightning functions and avoid recreating the wheels? 

2. For models, we support three modes, `kim`, `tar`, and `graph`. If I understand correctly, `tar` is just a local tarball of the `kim` mode. So can we merge these two? 

Detailed comments on these are provided in place.

kliff/trainer/kim_trainer.py
ipcamit: That is interesting, I could not find any obvious way of pickling Pybind11 C++ bindings using pickle. Can you provide more information on it? I dont think I saw it in KIM. Just for personal interest.

ipcamit:
I will go through the comments in coming week, as I get some time (currently working on making the Mac package for libdescriptor so all tests can pass). But regarding your major questions:

>    Can we merge torch_trainer.py and lightning_trainer.py, to use lightning functions and avoid recreating the wheels?

The reason for two separate trainers is that Lightning can get bit rigid at times, and there might be more complicated workflows. For example, EMA updates of weights needs explicitly uploading them to GPU, as Lightning did not have support for it yet. In such cases having a pure torch trainer might be useful. For large models, Lightning is essential, but for small, more innovative research issues, like say TorchMD-net etc, torch based trainer will be better on the side. 


> For models, we support three modes, kim, tar, and graph. If I understand correctly, tar is just a local tarball of the kim mode. So can we merge these two?

Yes, tar is just KIM model which is tarball. It can also be a ML model, in which case we will pluck out the torchscript wile and use that as model. So in that sense tar = KIM physics + KIM ML models. But you are right, I think I can merge them such that it will treat tarballs are ML or KIM model based on files inside.

kliff/dataset/dataset.py
```diff
@@ -760,23 +786,32 @@ def add_from_path(
...
         """
         if isinstance(path, str):
             path = Path(path)
-        configs = self._read_from_path(path, weight, file_format)
+
+        if isinstance(weight, Weight):
+            configs = self._read_from_path(path, weight, file_format)
+        else:
+            configs = self._read_from_path(path, None, file_format)
+            self.add_weights(weight)
```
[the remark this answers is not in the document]
ipcamit: Thanks. That one slipped past

kliff/models/kim.py
```diff
@@ -21,6 +26,13 @@
 except ImportError:
     kimpy_avail = False
 
+# list of model drivers that are not supported by this trainer.
+# example quip, torchml, etc.
+# TODO: Get the complete list of unsupported model drivers.
+UNSUPPORTED_MODEL_DRIVERS = [
```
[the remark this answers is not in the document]
ipcamit: Let me try, supported list might be too long

kliff/models/kim.py
```diff
@@ -681,6 +695,225 @@ def __call__(
...
+              Ask Mingjian/Yaser for comment.
+
+        Args:
+            model_name: name of the model.
+
+        Returns:
+            Model driver name.
+        """
+        # check if model is tarball
+        if "tar" in model_name:
```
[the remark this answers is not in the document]
ipcamit: No. tar is keyword in input file. If "tar" is provided in input, kliff will use tarfile module to untar it, irrespective of filename.

kliff/models/kim.py
```diff
@@ -681,6 +695,225 @@ def __call__(
...
+        else:
+            raise KIMModelError(f"Model type {model_type} not supported.")
+
+        model = KIMModel(model_name)
+
+        if param_manifest:
+            mutable_param_list = []
+            for param_to_transform in param_manifest.get("parameter", []):
+                if isinstance(param_to_transform, dict):
+                    parameter_name = list(param_to_transform.keys())[0]
```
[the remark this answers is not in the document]
ipcamit: I dont understand the question completely, if the parameter is a dict, then that dict will contain only single field, that is the name of the parameter, and its value will be the transform params.

Here is an example, where `A` and `B` are non-transformed, whereas `sigma` is.

```
['A',
 'B',
 {'sigma': {'transform_name': 'LogParameterTransform',
   'value': 2.0,
   'bounds': [[1.0, 10.0]]}}]
```
so `list(param_to_transform.keys())[0]` will give us `sigma`.

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+        training_manifest: training manifest
+    """
+
+    def __init__(self, training_manifest: dict, model=None):
+        # workspace variables
+        self.workspace: dict = {
+            "name": "kliff_workspace",
+            "seed": 12345,
+            "resume": False,
+            "walltime": "2:00:00:00",
```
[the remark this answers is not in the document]
ipcamit: OK. that will also make it simpler and easier.

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+        # part of current?
+        self.dataset_sample_manifest: dict = {
+            "train_size": None,
+            "val_size": None,
+            "indices_files": {"train": None, "val": None},
+            "val_indices": None,
+            "train_indices": None,
+        }
+        self.train_dataset = None
+        self.val_dataset = None
```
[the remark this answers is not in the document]
ipcamit: I did add it initially, but removed it as I did not see any use of it in MLIPs etc. I could not see that much of distinction between test and val sets. Can you tell where are you using `test` and `val` sets in you research? So that I can have some context

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+            "epoch": 0,
+            "step": 0,
+            "device": "cpu",
+            "expected_end_time": None,
+            "warned_once": False,
+            "dataset_hash": None,
+            "data_dir": None,
+            "appending_to_previous_run": False,
+            "verbose": False,
+            "ckpt_interval": 100,
```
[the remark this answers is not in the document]
ipcamit: saving top N models is actually useful. Will add it in future if time permits, else post release. It is trivial to save N models, but keeping ti sorted and moving model files on hard drive needs some thought.

kliff/trainer/kim_residuals.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Dict
+
+import numpy as np
+
+
+def MSE_residuals(
```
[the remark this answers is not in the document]
ipcamit: Leaving it unresolved for later discussion

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+                "kwargs": None,
+            },
+        }
+        self.property_transforms = []
+        self.configuration_transform = None
+
+        # training variables
+        # this is too complicated to put it in singe dict, therefore the training
+        # block is divided into loss, optimizer, dataset_sample
+        self.training_manifest: dict = {}
```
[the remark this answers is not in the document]
ipcamit: I dont understand, parse manifest divides the `training_manifest` in loss optimizer etc. Also training manifest also contains information that cannot be categorized so neatly, so it can be used. It is just that loss_manifest etc provides easy abstractions to categorize these data, as opposed to a single monolith dictionary. For example, exponential averaging of weights, can be categorized as model, or optimizer, or training manifest, so for such information `training_manifest` can be used.

kliff/trainer/base_trainer.py
```diff
@@ -0,0 +1,636 @@
...
+        self.current["device"] = self.training_manifest.get("device", "cpu")
+
+        # dataset sample variables will be processed in the setup_dataset method
+        self.export_manifest |= manifest.get("export", {})
+
+    def config_to_dict(self):
+        """
+        Convert the configuration to a dictionary.
+        """
+        config = {}
```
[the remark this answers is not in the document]
ipcamit: Corrected. It was an oversight. `to_dict` was supposed to recreate the input file

kliff/trainer/base_trainer.py
ipcamit: I have gone back and forth on this one several times. the reason why config transform is not applied here is that for large datasets like MPTraj, you run of memory really fast. So you evaluate transforms at model train time. For smaller datasets it is actually more useful to transform the model and save the binary for re-use. I think best would be to add a keyword `lazy_eval` etc. But it is much simpler this way that every configuration would be evaluated at runtime.

kliff/trainer/base_trainer.py
ipcamit: Renamed. For test, see comment above.

kliff/trainer/base_trainer.py
ipcamit: Agree. Removed shuffling.

kliff/trainer/torch_trainer.py
ipcamit: We can, but the real reason behind `torch_trainer` is to provide access to descriptor based models, and more flexible trainer for further custom training. Lightning trainer, which most will use, is limited to GNNs as 1) they are clear favorite and most sought after, 2) they can be trained on a scale where lightning makes a difference.

ipcamit:
I have added some changes as requested. Some are pending for next iteration (along with additions to torch lightning). I would like to discuss any changes regarding the API and design. As if that is finalized, I can start cleaning up the code and making it more consistent.

ipcamit:
Added fresh PR as per discussed. Only containing the lightning trainer for now. Will add KIM next. Removing this PR

PR #185, opened 2024-07-05
Kliff DNN torch trainer

ipcamit:
## Summary

Added tests and trainer for training generic dense neural networks using libdescriptor and pytorch (as opposed to lightning).

* Batched descriptor evaluation
* DNN model using TorchML driver

## TODO (if any)
* Need to add the capability to write the trained model as a valid DUNN driver model.

## Checklist

Before a pull request can be merged, the following items must be checked:

* [x] Make sure your code is properly formatted. [isort](https://pycqa.github.io/isort/) and [black](https://black.readthedocs.io/en/stable/getting_started.html) are used for this purpose. The simplest way is to use [pre-commit](https://pre-commit.com). See instructions [here](https://github.com/openkim/kliff/blob/main/docs/source/contributing_guide.md#code-style).
* [x] Doc strings have been added in the [Google docstring format](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) on your code.
* [x] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org) to type check your code.
* [x] Tests have been added for any new functionality or bug fixes.
* [x] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more efficient if you already fix most errors prior to submitting the PR.


kliff/trainer/torch_trainer_utils/dataloaders.py
```diff
@@ -82,33 +185,22 @@ def collate(self, batch: Any) -> dict:
         """
         # get fingerprint and consistent properties
         config_0, property_dict_0 = batch[0]
-        device = config_0.device
-        ptr = torch.tensor([0], dtype=torch.int64, device=device)
+        ptr = np.array([0], dtype=np.intc)
```
mjwen, line 188: I think the collated data will be provided as the input to a torch NN model. Any reason why use `np.array` but `torch.tensor` for all variables?

kliff/trainer/torch_trainer.py
```diff
@@ -0,0 +1,684 @@
...
+            the model must be a torchscript model, or valid TorchML model (tar or dir).
+    """
+
+    def __init__(self, configuration: dict, model=None):
+        self.configuration_transform: "Descriptor" = (
+            None  # for type checking the functions
+        )
+        self.torchscript_file = None
+        self.train_dataloader = None
+        self.validation_dataloader = None
```
mjwen, line 41: Do we want to add test_dataloader and test the model at the end of training?

mjwen (approved):

kliff/trainer/torch_trainer_utils/dataloaders.py
mjwen: Update: NVM, now I understand that the batched data are passed to the descriptor in `def _descriptor_eval_batch(self, batch) -> torch.Tensor`, which requires numpy array.

mjwen:
Looks great! Merged.

kliff/trainer/torch_trainer.py
ipcamit: Yes, I will add it uniformly across all trainers after finalizing this release. Idea is to have a base method `test()` that can run some tests on the test datasets. It will do simple energy and forces test but will not be limited to it, but could also leverage openkim tests if user requests.

Issue #203, opened 2025-03-05
Fitting certain parameters in Tersoff potential returns RuntimeError: the model does not provide an Extension routine


dengtq:
When fitting certain parameters in Tersoff potential from openkim, the following RuntimeError is returned:

```
File ***/.conda/envs/kliff/lib/python3.9/site-packages/kliff/models/kim.py", line 352, in compute
    raise kimpy.KimPyError("Calling `kim_model.compute()` failed.")
kimpy.err.KimPyError:
ERROR(@compute): Calling `kim_model.compute()` failed.
Traceback (most recent call last):
  File "***/.conda/envs/kliff/lib/python3.9/site-packages/kliff/models/kim.py", line 350, in compute
    kim_model.compute(self.kim_ca)
RuntimeError: the model does not provide an Extension routine!  or
the model's Extension routine returns error!
```

What is the cause of this error? Does that mean certain parameters cannot be optimized using kliff? Thank you for kind reply.

Environment:
Linux+Conda
kim-api  2.3.0
kimpy 2.1.1
kliff 0.4.3

Tersoff potential from openkim: Tersoff_LAMMPS_Tersoff_1989_SiC__MO_171585019474_004

mjwen:
@ilia-nikiforov-umn @ellio167 any idea? 

Does this mean the Tersoff driver does not expose parameters?

dengtq:
I'd clarify that this only happens for some of the parameters (Rc and Dc for example), while optimization of some other parameters like A and B works normally.

ellio167:
This error message comes from here (in

dengtq:
Thank you @ellio167 for the suggestion!

I have tested all parameters in Tersoff potential and looked into the kim.log file of failed ones.

For `beta`, the kim.log mentioned that `Parameter beta (Si-Si) may not be smaller than zero.`. So this is due to the parameter constraint and is resolved after I transform `beta` into logscale by defining a custom `ParameterTransform` class.

The case for `Rc` and `Dc` are different though. They return the following error:
```
2025-03-06:12:09:27CST * 0 * error * 0x5564613aff90_0x5564612ff790 * KIM_ComputeArgumentsImplementation.cpp:886 * Simulator supplied GetNeighborList() routine returned error.
2025-03-06:12:09:27CST * 0 * error * 0x5564613aff90 * model_driver_Tersoff.cpp:211 * compute: Error in KIM::ModelComputeArguments.GetNeighborList
[...]

PR #205, opened 2025-03-27
Debug cutoff radius update in KIM model

dengtq:
Partially address issue #203 .

The `ParameterTransform` problem is not resolved yet.


mjwen:
Thanks a lot!

PR #206, opened 2025-03-31
Kliff master v1

ipcamit:
- Removed 3.8 support. python 3.8 does not support all the typing hints.
- converted all `torch.load` commands to `torch.load(, weights_only=False)`. Newer pytorch versions give error.
- Corrections to tests
- Corrected weights broadcasting in the DNNTrainer (batching now also works with scalar weights)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -60,7 +60,7 @@ jobs:
 
           # install torch dependencies
           # Need to install torch manually as the install group cannot find torch
-          python -m pip install torch --index-url https://download.pytorch.org/whl/cpu 
+          python -m pip install torch --index-url https://download.pytorch.org/whl/cpu
           python -m pip install .[torch]
 
           # TODO, here, we install ptemcee from Yonatan's fork. See setup.py for details.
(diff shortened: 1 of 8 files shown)


ipcamit:
This is strange. Tests passes fine when I give just 3.11, but with all 3, suddenly it fails. Let me investigate it a bit

ipcamit:
All checks pass now. Issue was that without version specification in conda, it was pulling an older broken build (it prefers conda file extension over tar). Fixed now. 
Restored README to original version as well.

ipcamit:
@mjwen Just a reminder regarding the CI merge

ipcamit:
@mjwen What else is left for it to release?

mjwen:
I think it is almost there. 

The only thing I can think of is merging the branch `v1` to `master`. This might not be easy since both branches have been kept developed at the same time (well, for the master branch, the development are mainly bug fixes and such). 

I think we might want to proceed in the below way:
1. Create a new branch based on the current master and call it `legacy` and such.  
2. 
2.1 Merge `v1` into `master` (this might be difficult to do using the merge functionality offered by git)
2.2 Simply replace `master` by `v1`, and manually diff the files and copy the changes/fixes we want from the original `master`

We can pick one of 2.1 and 2.2. I have no preference. What do you think?

ipcamit:
I can give 2.1 a chance today, and then we can discuss accordingly.

PR #218, opened 2025-05-17
API Docs fix

ipcamit:
- Corrected small mistakes in the docs-
- Added rst files for some of the older notebooks (except UQ stuff)
- Fixed conf.py to generate api docs (lightning trainer docs does not work, but thats todo later)

Issue was a breaking change in sphinx. Earlier it used to generate `kliff.module.rst.` filenames there fore you had 4 list elements. Now there are just 3. Chatgpt suggested a more pythonic fix, that finally made into the conf.py file.


ipcamit:
Example: https://kliff-v1.readthedocs.io/en/latest/apidoc/kliff.html

ipcamit:
@mjwen Sorry just tagging in case this skipped your feed.

docs/source/conf.py
```diff
@@ -316,7 +332,7 @@ def autodoc_module(path: Path, module: str):
         f.write("    :members:\n")
         f.write("    :undoc-members:\n")
         # f.write("    :show-inheritance:\n")
-        f.write("    :inherited-members:\n")
+        # f.write("    :inherited-members:\n")
```
mjwen, line 335: Can you please remind me what `inherited-members` is, and why we want to remove it now?

mjwen (commented):
Looks great! Just one question. See the comments.

mjwen:
> @mjwen Sorry just tagging in case this skipped your feed.

Sorry, forgot this indeed.

docs/source/conf.py
ipcamit: It also starts documenting the base class members. For example if I uncomment it we also get base torch function like `register_hooks`, `parameter`, `children` in api docs. When we remove it it only catalogues the functions we implement.

ipcamit:
No issues, I have already prepared two more PRs for ASE 3.23 and some features to KLIFF graphs!

docs/source/conf.py
mjwen: Right. Now I remember that I added this to show the methods of a parent class in child classes for self-defined classes. Now that we have inherited many Torch classes, it is reasonable to not show stuff like `register_hooks`.

